PR #2, opened 2020-07-19
Reworked remote connection code

jairinhohw:
I've tried to use the local connection used in the existing tests, but could not make them work (keeps giving "Your user name and password are not defined"), so i tried to test the remote connection code and it crashed with a double free while freeing the dpb (i don't really know why exactly).

So i modified the code to be more similar to the one i made while testing on my own an it is connecting now.

I'm thinking in refactoring the rest of the code as i can to use more safe rust code instead of raw pointers and mallocs if you are ok with it, as i did in the connection code.


fernandobatels:
> I've tried to use the local connection used in the existing tests, but could not make them work (keeps giving "Your user name and password are not defined").

What is the version of your firebird client? 3.0? 

>  so i tried to test the remote connection code and it crashed with a double free while freeing the dpb (i don't really know why exactly).

I had a similar experience with crashes when I re-use the same transaction for more that one query. But the remote connections works fine to me.

> So i modified the code to be more similar to the one i made while testing on my own an it is connecting now.

Wonderful, in my tests with firebird 3.0 your fixes are also working.

> I'm thinking in refactoring the rest of the code as i can to use more safe rust code instead of raw pointers and mallocs if you are ok with it, as i did in the connection code.

Of course. I'm open to any change in the original code. I don't have much experience with the firebird c API, so, my code use lot of unsafe calls.

You are ok with the "api structure" of connections, transactions and querys? You think that can be improved or changed? 

I will make a little comment in your pull request, but your changes are very welcome!

src/connection.rs
```diff
@@ -173,14 +181,26 @@ mod test {
...
-            .expect("error on close the connection");
+        conn.close().expect("error on close the connection");
+    }
+
+    #[test]
+    fn remote_connection() {
+        let conn = Connection::open(
+            "localhost".into(),
+            3050,
+            "/var/lib/firebird/system/test.fdb".into(),
```
fernandobatels, line 198: I think we can't use a full path file into tests or examples.

We can use paths like "test.fdb" with the "DatabaseAccess = Restrict" setting into firebird.conf(https://firebirdsql.org/rlsnotesh/config-fb-conf.html)

jairinhohw: Oh, thanks for the info, i will fix it

jairinhohw:
> What is the version of your firebird client? 3.0?

I've tested with firebird server and lib version 2.5.9.

> You are ok with the "api structure" of connections, transactions and querys? You think that can be improved or changed?

I'll think of something, right now i think its more important to fix the invalid memory reads i found running the tests through valgrind.

As a side note, is it ok to format the code with `cargo fmt`? My editor is configured to run it automatically, so i've accidentally reformatted the code before commiting, but i think its better to use `cargo fmt` to make the formatting consistent.

fernandobatels:
> I'll think of something, right now i think its more important to fix the invalid memory reads i found running the tests through valgrind.

Got it

> As a side note, is it ok to format the code with `cargo fmt`? My editor is configured to run it automatically, so i've accidentally reformatted the code before commiting, but i think its better to use `cargo fmt` to make the formatting consistent.

My fault. I will apply the cargo fmt in the project right now and we will use it.

jairinhohw:
Another side note: To document a module from within (the doc comment on top of the file) you must use `//!` instead of `///`. [link]

[link]: https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#commenting-contained-items

fernandobatels:
> Another side note: To document a module from within (the doc comment on top of the file) you must use `//!` instead of `///`. [link](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#commenting-contained-items)

Fixed

PR #4, opened 2020-07-20
Minimized use of unsafe and a bit of changes to the api

jairinhohw:
I've also changed some of the types in the `ibase` module back to the originals generated by bindgen, i think it's better
to not change the autogenerated code to simplify moving to newer versions of the fbclient, and the module
could be autogenerated if we can make static linking work, what do you think?

### DIFF
--- a/src/ibase.rs
+++ b/src/ibase.rs
@@ -2445,7 +2445,7 @@ pub type PARAMDSC = paramdsc;
 #[derive(Debug, Copy, Clone)]
 pub struct paramvary {
     pub vary_length: ISC_USHORT,
-    pub vary_string: [ISC_UCHAR; 32usize],
+    pub vary_string: [ISC_UCHAR; 1usize],
 }
 pub type PARAMVARY = paramvary;
 #[repr(C)]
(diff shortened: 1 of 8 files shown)


src/ibase.rs
```diff
@@ -2474,11 +2474,11 @@ pub struct XSQLDA {
     pub sqldabc: ISC_LONG,
     pub sqln: ISC_SHORT,
     pub sqld: ISC_SHORT,
-    pub sqlvar: [XSQLVAR; 32usize],
+    pub sqlvar: [XSQLVAR; 1usize],
```
jairinhohw, line 2477: This change was making the length calculation alloc more than necessary, because mem::size_of<XSQLDA>() was returning a lenght of a XSQLDA with space for 32 fields of XSQLVAR instead of just one as the c macro implied.

fernandobatels:
> I've also changed some of the types in the `ibase` module back to the originals generated by bindgen, i think it's better
> to not change the autogenerated code to simplify moving to newer versions of the fbclient, and the module
> could be autogenerated if we can make static linking work, what do you think?

I agree. I will make a small changes in the error.rs and the examples, but the PR are OK and we can accept then

PR #12, opened 2020-07-23
Fixed wrong BIGINT buffering. (Fix #11)

### LINKED ISSUE #11: Failed to map BIGINT using i64
Hi.

Consider the following table structure and its data:

```sql
CREATE TABLE test (
  id BIGINT GENERATED BY DEFAULT AS IDENTITY NOT NULL PRIMARY KEY,
  name VARCHAR(100) NOT NULL UNIQUE
);

INSERT INTO test (name) VALUES ('Fulano');
INSERT INTO test (name) VALUES ('Beltrano');
INSERT INTO test (name) VALUES ('Cicrano');

SELECT * FROM test;

-- ID	NAME
-- 1	Fulano
-- 2	Beltrano
-- 3	Cicrano
```

Now, consider the following example:

```rust
use rsfbclient::{Connection, FbError};

fn main() -> Result<(), FbError> {
    let conn = Connection::open("localhost", 3050, "test.fdb", "SYSDBA", "masterkey")?;
    let tr = conn.transaction()?;
    let rows = tr.prepare("SELECT * FROM test")?.query(())?.into_iter();
    println!("| ID | NAME |");
    println!("| -- | ---- |");
    for row in rows {
        let (id, name): (i64, String) = row?;
        println!("| {} | {} |", id, name);
    }
    Ok(())
}
```

It is returning the following result:

```bash
| ID | NAME |
| -- | ---- |
| 4294967297 | Fulano |  
| 8589934594 | Beltrano |
| 12884901891 | Cicrano |
```

but it should return:

```bash
| ID | NAME |
| -- | ---- |   
| 1 | Fulano |  
| 2 | Beltrano |
| 3 | Cicrano |
```

regards,

P.S.: Changing the column type from `i64` to `i32` it prints the values properly for short ones, however, it isn't managed to map a full bigint range (-2^63 .. 2^63-1).

silvioprog:
Related to https://github.com/fernandobatels/rsfbclient/issues/11.

### DIFF
--- a/src/row.rs
+++ b/src/row.rs
@@ -304,7 +304,7 @@ fn integer_from_buffer(buffer: &[u8]) -> Result<i64, FbError> {
     }
 
     Ok(i64::from_le_bytes([
-        buffer[0], buffer[1], buffer[2], buffer[3], buffer[0], buffer[1], buffer[2], buffer[3],
+        buffer[0], buffer[1], buffer[2], buffer[3], buffer[4], buffer[5], buffer[6], buffer[7],
     ]))
 }
 


fernandobatels:
Thanks for the fix!

PR #19, opened 2020-07-28
Higher level api, Statement Cache and r2d2 Connection Pool

jairinhohw:
Added the `Queryable` trait to provide a higher level api that abstracts over the transaction creation / preparing / executing / collecting the results.

Added the `StmtCache` for caching the prepared statements and reusing them while using the higher level api.

Added the `FirebirdConnectionManager` used to build an `r2d2` connection pool.


src/connection/pool.rs
```diff
@@ -0,0 +1,35 @@
...
+    type Error = FbError;
+
+    fn connect(&self) -> Result<Self::Connection, Self::Error> {
+        self.conn_builder.connect()
+    }
+
+    fn is_valid(&self, conn: &mut Self::Connection) -> Result<(), Self::Error> {
+        let mut tr = Transaction::new(&conn)?;
+        tr.execute_immediate("SELECT 1 FROM RDB$DATABASE", ())
+    }
```
fernandobatels, line 30: I my tests, i caught the following error, running the pool example: 

```get connection from pool error in line:51 ! error: Error(Some("-902: Dynamic SQL Error\nSQLDA error\nNo SQLDA for output values provided"))```

This occurs because the utilization of "tr.execute_immediate" in a query that return rows. When I a change to use the "prepare and query" the example works fine:

```
Thread 0: 0.8083343538681508
Thread 2: 0.6999827182100219
Thread 1: 0.9977548040080352
```

I'm using the Firebird 3.0.4 in my tests.

Unfortunately, I did not find any explanation about a change in the SQLDA on firebird 3.

PR #33, opened 2020-08-13
Blob

fernandobatels:


rsfbclient-native/src/row.rs
```diff
@@ -124,12 +142,75 @@ impl ColumnBuffer {
...
+    db: &mut ibase::isc_db_handle,
+    tr: &mut ibase::isc_tr_handle,
+    ibase: &IBase,
+) -> Result<String, FbError> {
+    let mut final_string = String::new();
+
+    let mut status = Status::default();
+    let mut handle = 0;
+
+    let (head, body, _) = unsafe { buffer.align_to::<ibase::GDS_QUAD_t>() };
```
jairinhohw, line 165: I think something on the lines of the `timestamp_from_buffer` can be used here to avoid the unsafe and assertion.

Related code:
```rust
let date = ibase::ISC_TIMESTAMP {
    timestamp_date: ibase::ISC_DATE::from_ne_bytes(buffer[0..4].try_into().unwrap()),
    timestamp_time: ibase::ISC_TIME::from_ne_bytes(buffer[4..8].try_into().unwrap()),
};
```

fernandobatels: Done!

rsfbclient-native/src/params.rs
```diff
@@ -123,9 +137,66 @@ impl ParamBuffer {
...
+fn binary_to_blob(
+    buffer: Vec<u8>,
+    db_handle: &mut ibase::isc_db_handle,
+    tr_handle: &mut ibase::isc_tr_handle,
+    ibase: &IBase,
+) -> Result<Vec<u8>, FbError> {
+    let mut status = Status::default();
+    let mut handle = 0;
+
+    let mut blob_id = MaybeUninit::<ibase::GDS_QUAD_t>::uninit();
```
jairinhohw, line 157: You can just initialize with 0 instead of using `MaybeUninit`.

```suggestion
    let mut blob_id = ibase::GDS_QUAD_t { gds_quad_high: 0, gds_quad_low: 0 };
```

rsfbclient-native/src/params.rs
```diff
@@ -123,9 +137,66 @@ impl ParamBuffer {
...
+    debug_assert_ne!(handle, 0);
+
+    let buffer_i8: Vec<i8> = buffer.iter().map(|byte| *byte as i8).collect();
+
+    unsafe {
+        if ibase.isc_put_segment()(
+            &mut status[0],
+            &mut handle,
+            buffer_i8.len() as u16,
+            buffer_i8.as_ptr(),
```
jairinhohw, line 182: The `buffer_i8` is not needed.
```suggestion
    unsafe {
        if ibase.isc_put_segment()(
            &mut status[0],
            &mut handle,
            buffer.len() as u16,
            buffer.as_ptr() as *mut i8,
```

fernandobatels: Done

rsfbclient-native/src/params.rs
fernandobatels: Done

PR #31, opened 2020-08-11
Run tests with fb v2 v3 and v4

juarezr:
## Changes

Continuous integration in Github now tests:

- building, clippy, code formatting
- FirebirdSQL versions 2.5-ss 3.0 4.0
- linking, dynamic_loading
- client, embedded
- coverage

This is a follow up to #25.

PR #34, opened 2020-08-16
ci: run tests on windows

juarezr:
### ci: run tests on windows

Continued from #31.

What's missing yet:

- `cargo tarpaulin` is disabled because it's failing on some path issue.
- `embedded` test case is not enabled.


juarezr:
> The `cargo tests` and `cargo tarpaulin` should use the `--no-default-features` flags, because of the `linking` feature is in the default set, which makes the `linking` tests run on all test variants.

It looks like this would be easier handled when we have [rust-lang#2911](https://github.com/rust-lang/cargo/issues/2911).

Issue #25, opened 2020-08-07
Continuing the GitHub Actions setup


fernandobatels:
* [ ] Next steps
  * [ ] Run test with Firebird 2.5-ss, 3.0 and 4.0
  * [ ] Run on Windows
  * [ ] Run on Linux/ARM
  * [ ] Run on Linux/ARM/LittleEndian
  * [ ] Release workflow to crates.io and rust docs.
* [ ] Status
  * [x] Coverage failing on cc linking (needs help...)
  * [ ] Next steps on further PRs
-------
Continuation of #24

fernandobatels:
For you @juarezr ;)

juarezr:
Any clue about this [failure](https://github.com/fernandobatels/rsfbclient/runs/959232114?check_suite_focus=true) ?

Anybody can help me figuring out what's failing in the build?

jairinhohw:
I couldn't find out the problem, but for now you can build with the `dynamic_loading` for it to work

PR #35, opened 2020-08-18
ci: run tests on macos-latest

juarezr:
Using github actions  with macos-latest configuration.

Similar to #34.
Follow-up to #25

### DIFF
--- a/rsfbclient-native/build.rs
+++ b/rsfbclient-native/build.rs
@@ -5,10 +5,16 @@ fn main() {
         #[cfg(all(feature = "linking", target_os = "linux"))]
         search_on_linux();
 
+        #[cfg(all(feature = "linking", target_os = "macos"))]
+        search_on_macos();
+
         #[cfg(all(feature = "linking", target_os = "windows"))]
         search_on_windows();
     }
+    println!("cargo:rerun-if-env-changed=PROFILE");
+    println!("cargo:rerun-if-env-changed=TARGET");
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=src/tests/mod.rs");
 }
 
 fn search_on_environment_var() -> bool {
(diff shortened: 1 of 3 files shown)


rsfbclient-native/build.rs
```diff
@@ -28,6 +34,26 @@ fn search_on_linux() {
...
+    }
+    let def_fbclient_lib =
+        "/Library/Frameworks/Firebird.framework/Versions/A/Libraries/libfbclient.dylib";
+    let fb3_lib_path_lib = std::path::Path::new(def_fbclient_lib);
+    if fb3_lib_path_lib.exists() {
+        println!(
+            "cargo:rustc-link-search=/Library/Frameworks/Firebird.framework/Versions/A/Libraries/"
+        );
+    }
+    println!("cargo:rustc-link-lib=dylib=libfbclient");
```
jairinhohw, line 52:
```suggestion
    println!("cargo:rustc-link-lib=dylib=fbclient");
```

juarezr: I'm unsure about this change because:
1. `cargo build` succeeds. 
2. `cargo test` fails with the same arguments.
3. The steps are in sequence. 

I have thinking the following hipotesis:
1. If `libfbclient` is wrong, `cargo build` may have failed before.
2. For some reason `cargo test` is not getting the search path of build. rs

What I am missing?

juarezr: Also this PR has the `cargo test` of the feature `linking` on `macos-latest` disabled for not breaking the build.
So we can solve this later.

jairinhohw: Build passes because only when building a executable (tests) the compiler call the linker, which passes the compiler flag `-llibrsfbclient` with `cargo:rustc-link-lib=dylib=libfbclient`. 
However, the linux version with `cargo:rustc-link-lib=dylib=fbclient` calls the linker with `-lfbclient` and works, and testing with `cargo:rustc-link-lib=dylib=libfbclient`, it fails with the same error as the current commit of this PR. (`cannot find -llibfbclient`)

jairinhohw: We can call the cargo build with the `--all-targets` flag to make sure the tests and examples are built, allowing the linker to be called without needing to run the tests.

jairinhohw: I will approve the pr to fix this later then.

PR #37, opened 2020-08-25
Named columns

fernandobatels:
Now we can use the name/alias of fetched columns


rsfbclient-core/src/row.rs
```diff
@@ -36,13 +36,25 @@ impl Row {
...
+    pub value: Option<ColumnType>,
+    pub name: String,
+}
+
+impl Column {
+    pub fn new(name: String, value: Option<ColumnType>) -> Self {
+        Column {
+            name: name,
+            value: value,
+        }
```
jairinhohw, line 49:
```suggestion
        Column {
            name,
            value,
        }
```

To make clippy happy

rsfbclient-native/src/row.rs
```diff
@@ -130,10 +133,19 @@ impl ColumnBuffer {
 
         var.sqldata = buffer.as_mut_ptr() as _;
 
+        let col_name = {
+            let bname = var.aliasname.iter().map(|b| *b as u8).collect::<Vec<u8>>();
+
+            str::from_utf8(&bname)
+                .map(|str| str.to_string())
+                .map_err(|_| FbError::from("Found a column name with an invalid utf-8 string"))
+        }?;
```
jairinhohw, line 142:
```suggestion
        let col_name = {
            let len = usize::min(var.aliasname_length as usize, var.aliasname.len());
            let bname = var.aliasname[..len].iter().map(|b| *b as u8).collect();

            String::from_utf8(bname)
                .map(|str| str)
                .map_err(|_| FbError::from("Found a column name with an invalid utf-8 string"))
        }?;
```

It was copying null bytes too

examples/namedcolumns_select.rs
```diff
@@ -0,0 +1,53 @@
...
+        .user("SYSDBA")
+        .pass("masterkey")
+        .connect()?;
+
+    let rows: Vec<Row> = conn.query("select test.*, 10 as extra from test", ())?;
+
+    for row in rows {
+        println!("------------------------------------");
+
+        for col in row.get_all::<Row>()?.cols {
```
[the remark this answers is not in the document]
fernandobatels: Done

PR #39, opened 2020-08-28
Bugfixes

jairinhohw:
Fixed some bugs found by using large blobs


jairinhohw:
I have no idea why the `macos` with `dynamic_loading` is failing

juarezr:
> I have no idea why the `macos` with `dynamic_loading` is failing

Let me see what happened.

jairinhohw:
Sadly no change, and the test only writes alphanumeric chars, so it shouldn't be a utf8 problem, it seems to occur on a specific lengths actually (The blob segments are received in chunks of 256 bytes in the wire, and the `\u{3}` chars seems to be 256 bytes appart of each other).
Maybe it's a bug on the shared library, as the pure_rust test passes (the `linking` is not being tested at the moment, as we could not get it to link).

jairinhohw:
It was a problem with the `CString` use, as the segments are not `0` terminated

PR #47, opened 2020-09-06
Charset support added

### LINKED ISSUE #45: Encoding
How can I specify a encoding to connect to a database?

Example in python:

```
conn  = fdb.connect(
    user='SYSDBA',
    password='masterkey',
    database='/firebird/data/BASE.FDB',
    charset='win1252'
)
```

fernandobatels:
Support for charsets added. I added the test `column_with_charset_none` to cover the error reported in #45. By default, the lib still using the std::from_utf8


rsfbclient-native/src/connection.rs
```diff
@@ -127,20 +126,22 @@ impl FirebirdClient for NativeFbClient {
...
                 stmt_data_map: Default::default(),
+                charset,
             }),
 
             #[cfg(feature = "dynamic_loading")]
             Args::DynamicLoading { lib_path } => Ok(Self {
                 ibase: IBase::with_client(lib_path).map_err(|e| FbError::from(e.to_string()))?,
                 status: Default::default(),
                 stmt_data_map: Default::default(),
+                charset: charset,
```
jairinhohw, line 144:
```suggestion
                charset,
```
Clippy is not happy

rsfbclient-rust/src/client.rs
```diff
@@ -86,11 +97,14 @@ impl FirebirdClient for RustFbClient {
...
 
-    fn new(_args: Self::Args) -> Result<Self, FbError>
+    fn new(charset: Charset, _args: Self::Args) -> Result<Self, FbError>
     where
         Self: Sized,
     {
-        Ok(Self { conn: None })
+        Ok(Self {
+            conn: None,
+            charset: charset,
```
jairinhohw, line 106:
```suggestion
            charset,
```
Clippy is not happy

rsfbclient-rust/src/client.rs
```diff
@@ -299,6 +314,7 @@ impl FirebirdWireConnection {
             version,
             buff,
             stmt_data_map: Default::default(),
+            charset: charset.clone(),
```
jairinhohw, line 317:
```suggestion
            charset,
```
Clippy again

src/connection/mod.rs
```diff
@@ -228,9 +241,15 @@ where
         self
     }
 
+    /// Charset. Default: UTF_8
+    pub fn charset(&mut self, charset: Charset) -> &mut Self {
```
jairinhohw, line 245:
```suggestion
    /// Connection charset. It is only necessary to specify a charset other than the default `UTF-8` if you
    /// have text stored in the database using columns with charset `NONE` or `OCTETS`. Otherwise
    /// the database will handle the charset conversion automatically
    pub fn charset(&mut self, charset: Charset) -> &mut Self {
```
Just to tell that it is normally not needed to specify a charset.

src/tests/params.rs
```diff
@@ -9,6 +9,38 @@ mk_tests_default! {
...
+        assert_eq!("Pão de queijo", pao);
+
+        let mut conn = connect(); // utf8
+
+        let err: Result<Option<(String,)>, FbError> = conn.query_first("select * from PCHARSETS", ());
+        assert!(err.is_err());
+        assert_eq!("error: Found column with an invalid UTF-8 string: invalid utf-8 sequence of 1 bytes from index 1", err.err().unwrap().to_string());
+
+        Ok(())
+    }
```
jairinhohw, line 42: The charset tests only runs in the `linking` version. I thing it's better to crate a new module for the charset tests and return two connections, one `UTF8 ` encoded and one `ISO_8859_1` encoded using the `mk_tests!` macro.

src/tests/row.rs
```diff
@@ -11,6 +11,72 @@ mk_tests_default! {
...
+    #[cfg(feature = "linking")]
+    #[cfg(not(feature = "embedded_tests"))]
+    #[test]
+    fn column_with_charset_none() -> Result<(), FbError> {
+        use crate::{ConnectionBuilder, charset::ISO_8859_1};
+
+        // This test reproduce the of using a diferent
+        // charset of column in insert.
+        // When we read the data, even using the UTF8,
+        // we got an "invalid utf-8" error.
```
jairinhohw, line 23: Just a comment, why the "even using UTF8"? The data was `ISO8859_1` encoded, so
it makes sense to not be able to read it as `UTF8`.

jairinhohw (changes requested):

jairinhohw:
One more thing, you need to encode the `sql` strings too with the connection encoding, else it will not work with sqls containing
special characters. This is a good exemple to crate a test, like this:
```rust
let mut conn = ConnectionBuilder::linked()
    .charset(ISO_8859_1)
    .connect()
    .expect("Error on connect the test database using the ISO8859_1");
let (pao,) = conn.query_first("SELECT cast('pão de queijo' as Varchar(30)) FROM RDB$DATABASE;", ())?.unwrap();
assert_eq!("pão de queijo", pao);
``'

File: rsfbclient-core/src/charset.rs [hunk 7]
```diff
@@ -0,0 +1,192 @@
...
+pub const WIN_1258: Charset = Charset {
+    fb: "WIN1258",
+    str: Some(all::WINDOWS_1258),
+};
+
+/// English
+pub const ASCII: Charset = Charset {
+    fb: "ASCII",
+    str: Some(all::ASCII),
+};
```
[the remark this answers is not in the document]
fernandobatels: Hmm. Ok, but if the user input an utf8 string like "pão" into an ASCII column? Using the ASCII column, the lib will handle the error

src/tests/row.rs
fernandobatels: My mistake. Fixed.

src/tests/params.rs
fernandobatels: A new module for charset OK, but I did not understand the "return thow connections". You are talking about the 'connect()' function?

jairinhohw: Yes, you can change it to return a tuple with two connections for example

jairinhohw: Creating a copy of the macro defined in line 44 of the `tests/mod.rs` file, that returns two connections instead of one, to use in the charset tests, that was my idea.

rsfbclient-core/src/charset.rs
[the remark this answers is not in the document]
jairinhohw: Well, i don't really see anyone using an ASCII only column, but ok, i just though it was unnecessary.

src/tests/params.rs
fernandobatels: Now I get it, but i think that we can use a other approach. Instead of duplicate the mk_tests_default, we can return a ConnectionBuilder instance. This will provide a greater flexibility for ours tests.

fernandobatels:
@jairinhohw, when I run the [params charset test](https://github.com/fernandobatels/rsfbclient/blob/nonutf8/src/tests/charset.rs#L12), I receive this error:
`test tests::charset::pure_rust::params ... test tests::charset::pure_rust::params has been running for over 60 seconds`

I traced the code until the [read_packet function](https://github.com/fernandobatels/rsfbclient/blob/nonutf8/rsfbclient-rust/src/client.rs#L684). The `socker.read` are 'locked'.

You have any suggestions?

jairinhohw (changes requested):
I'm working on the problem with the pure_rust.

rsfbclient-native/src/params.rs
```diff
@@ -119,10 +122,13 @@ impl ParamBuffer {
 
         let mut buffer = match info {
             Param::Text(s) => {
-                if s.len() > MAX_TEXT_LENGTH {
-                    binary_to_blob(s.into_bytes(), db, tr, ibase)?
+                let len = s.len();
+                let bytes = charset.encode(s)?;
```
jairinhohw, line 126: The length is wrong, you need to use the length after encoding, not before.

```suggestion
                let bytes = charset.encode(s)?;
                let len = bytes.len();
```

rsfbclient-rust/src/blr.rs
```diff
@@ -59,17 +59,19 @@ pub fn params_to_blr(
     for p in params {
         match p {
             Param::Text(s) => {
-                if s.len() > MAX_DATA_LENGTH {
+                let len = s.len();
+                let bytes = conn.charset.encode(s.to_string())?;
```
jairinhohw, line 63: The length is wrong, you need to use the length after encoding, not before.
This is what was causing the code to hang, as it was sending a malformed message

```suggestion
                let bytes = conn.charset.encode(s.to_string())?;
                let len = bytes.len();
```

jairinhohw (changes requested):
Found the problem causing the `pure_rust` to hang.

rsfbclient-native/src/connection.rs
```diff
@@ -243,12 +243,15 @@ impl FirebirdClient for NativeFbClient {
         dialect: Dialect,
         sql: &str,
     ) -> Result<(), FbError> {
+        let len = sql.len();
+        let sql = unsafe { CString::from_vec_unchecked(self.charset.encode(sql)?) };
```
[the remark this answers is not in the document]
fernandobatels: Hmm. I don't realized that I could use byte array directly. I was using the CString for prevent the UTF8 conversion o String.

Fixed!

rsfbclient-native/src/connection.rs
```diff
@@ -267,6 +270,9 @@ impl FirebirdClient for NativeFbClient {
         dialect: Dialect,
         sql: &str,
     ) -> Result<(StmtType, Self::StmtHandle), FbError> {
+        let len = sql.len();
+        let sql = unsafe { CString::from_vec_unchecked(self.charset.encode(sql)?) };
```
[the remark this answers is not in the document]
fernandobatels: Hmm. I don't realized that I could use byte array directly. I was using the CString for prevent the UTF8 conversion o String.

Fixed!

rsfbclient-rust/src/wire.rs
```diff
@@ -241,18 +247,23 @@ pub fn transaction_operation(tr_handle: u32, op: TrOp) -> Bytes {
...
 
 /// Execute immediate request
-pub fn exec_immediate(tr_handle: u32, dialect: u32, sql: &str) -> Bytes {
+pub fn exec_immediate(
+    tr_handle: u32,
+    dialect: u32,
+    sql: &str,
+    charset: &Charset,
+) -> Result<Bytes, FbError> {
     let mut req = BytesMut::with_capacity(28 + sql.len());
```
jairinhohw, line 256: The capacity calculation is counting always the `UTF-8` byte count, so it may be wrong in other encodings.

rsfbclient-rust/src/wire.rs
```diff
@@ -267,19 +278,25 @@ pub fn allocate_statement(db_handle: u32) -> Bytes {
...
 /// in the previous request
-pub fn prepare_statement(tr_handle: u32, stmt_handle: u32, dialect: u32, query: &str) -> Bytes {
+pub fn prepare_statement(
+    tr_handle: u32,
+    stmt_handle: u32,
+    dialect: u32,
+    query: &str,
+    charset: &Charset,
+) -> Result<Bytes, FbError> {
     let mut req = BytesMut::with_capacity(28 + query.len() + XSQLDA_DESCRIBE_VARS.len());
```
jairinhohw, line 288: The capacity calculation is counting always the `UTF-8` byte count, so it may be wrong in other encodings.

rsfbclient-native/src/params.rs
fernandobatels: Fixed!

rsfbclient-rust/src/blr.rs
fernandobatels: Fixed!

rsfbclient-rust/src/wire.rs
fernandobatels: Fixed!

rsfbclient-rust/src/wire.rs
fernandobatels: Fixed!

fernandobatels:
> Found the problem causing the `pure_rust` to hang.

Hmm. I think no. When I run all tests, the wire implementation still locking

jairinhohw:
It worked only for the Firebird 2.0, so it must have another detail missing

jairinhohw (approved):

jairinhohw:
All tests run fine in my machine too

fernandobatels:
> All tests run fine in my machine too

This is stranger. In my machine the pure_rust still failing. But, when I run only the `tests::charset::pure_rust::params` work correctly

jairinhohw:
> > All tests run fine in my machine too
> 
> This is stranger. In my machine the pure_rust still failing. But, when I run only the `tests::charset::pure_rust::params` work correctly

Does it hang or cause an error?

fernandobatels:
> > > All tests run fine in my machine too
> > 
> > 
> > This is stranger. In my machine the pure_rust still failing. But, when I run only the `tests::charset::pure_rust::params` work correctly
> 
> Does it hang or cause an error?

Hang. I use a firebird 3. But if works in our CI its OK. I will wait for the @juarezr review and tests.

fernandobatels:
@mariuz, you can help us reviewing this PR?

rsfbclient-rust/src/blr.rs
```diff
@@ -59,17 +59,18 @@ pub fn params_to_blr(
     for p in params {
         match p {
             Param::Text(s) => {
-                if s.len() > MAX_DATA_LENGTH {
+                let bytes = conn.charset.encode(s.to_string())?;
```
juarezr, line 62: If I'm not wrong, here the call `s.to_string()` makes a copy of the content of variable `s` and `charset.encode()` does a second copy, right?

This could be replaced by something like:
``` rust
    for p in params {
        match p {
            Param::Text(&'a s) => {
                if s.len() > MAX_DATA_LENGTH {
                let bytes = conn.charset.encode<'a>(&s)?;
```

I have the impression that code in this place will be very hot and optimizing or helping the compiler optimize will contribute for raw performance.

rsfbclient-core/src/charset.rs
```diff
@@ -0,0 +1,194 @@
...
+            })
+        } else {
+            str::from_utf8(bytes)
+                .map(|str| str.to_string())
+                .map_err(|e| format!("Found column with an invalid UTF-8 string: {}", e).into())
+        }
+    }
+
+    // Encode the string into bytes using the current charset
+    pub fn encode<S: Into<String>>(&self, str: S) -> Result<Vec<u8>, FbError> {
```
juarezr, line 37: The trait `Into<T>` here and the corresponding `&str.into()` are very handy for use, but theyt incur in a small penalty because of the dynamic dispatch that they use.

Maybe we could have a overloaded/alternative function using `&str` that could be used at leas in internal code for high performance?

``` rust
pub fn encode_str(&self, str: &str) -> Result<Vec<u8>, FbError> {
    // ...
}
```

rsfbclient-core/src/charset.rs
```diff
@@ -0,0 +1,194 @@
...
+use encoding::{all, types::EncodingRef, DecoderTrap, EncoderTrap};
+use std::str;
+
+use crate::FbError;
+
+/// Charset definition. Used to encode/decode the
+/// strings.
+pub struct Charset {
+    pub fb: &'static str,
+    pub str: Option<EncodingRef>,
```
juarezr, line 14: The field name `str` used in `pub str: Option<EncodingRef>` clashes with the type name `str` as in `pub fb: &'static str`.

Maybe rename for avoiding confusion or future problems with new Clippy lints, compiler warnings ?

rsfbclient-core/src/connection.rs
```diff
@@ -32,7 +32,7 @@ pub trait FirebirdClient: Send {
     /// Arguments to instantiate the client
     type Args: Send + Sync + Clone;
 
-    fn new(args: Self::Args) -> Result<Self, FbError>
+    fn new(charset: Charset, args: Self::Args) -> Result<Self, FbError>
```
juarezr, line 35: I don't remember that is possible to retrieve/switch automatically the `dialect` and `charset` from the defaults in the connecting database.

Doing this could be handy for new users that don't have this concepts well understood.

Maybe having a variant/alternative `connect()` for this case would be good.

Of course any query that should be executed on connection time can hurt applications that don't use connection pooling and don't have long lived connections. This will be common in applications like little websites and microservice. For this case have the option of setting `dialect` and `charset` would optimize.

juarezr (changes requested):
I'm submitting feedback, but don't know if all point are a valid concern. Also they could be implemented in further PRs if anyone would make sense.

rsfbclient-rust/src/blr.rs
jairinhohw: In this case, just removing the `to_string()` call will fix the unnecessary copy.

We should not change the encode to take a reference, because in the present case, we pass a owned string,
and if the encoding is utf8, we can just return the passed string underlying vector, without making any copy,
however, when the encoding is not utf8, we need to make a copy, as the bytes will likely be different than the ones
in the original string.

rsfbclient-core/src/charset.rs
jairinhohw: While i do agree that the `Into<String>` here is not necessary, rust generics do not cause dynamic dispatch, they work by creating multiple versions of the function for the necessary types, so it has only compiletime overhead.
Dynamic dispatch would be used if the type of the str variable was `Box<dyn Into<String>>` for example.

juarezr: Ok, rust monomorphization will handle all this during the compiler phase.

rsfbclient-rust/src/blr.rs
juarezr: I didn't understand exactly if you mean:

 - [x] remove `to_string()`
- [ ] change `charset.encode()` signature

jairinhohw: This:
```suggestion
                let bytes = conn.charset.encode(s)?;
```

src/tests/charset.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    #[test]
+    fn cast_charsets() -> Result<(), FbError> {
+        let mut conn = cbuilder().connect()?;
+
+        let qsql = "select
+                            CAST('olá abc ë' AS VARCHAR(10) CHARACTER SET WIN_1252),
+                            CAST('olá abc ë' AS VARCHAR(10) CHARACTER SET ISO8859_1),
+                            CAST('olá abc ë' AS VARCHAR(10) CHARACTER SET utf8),
+                            CAST('olá abc ë' AS VARCHAR(11) CHARACTER SET none)
```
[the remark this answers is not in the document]
fernandobatels: I think that we don't need assert all languages, because the UTF8 will be de same. I will keep only our Portuguese characters for now.

rsfbclient-rust/src/blr.rs
fernandobatels: Good catch. Fixed!

This 'to_string()' remained of an old encode method version.

rsfbclient-core/src/connection.rs
fernandobatels: Don't worry, the users can't access this method. We only use this internally.

rsfbclient-core/src/charset.rs
fernandobatels: Done

jairinhohw (approved):

juarezr (approved):
Nice work!

PR #50, opened 2020-09-13
Refactorations

jairinhohw:
In the first commit, changes the encoding and decoding code to avoid copying data where possible by using the `Cow` type.

In the second, change the way data is sent / received from the `fbclient` library to make sure all data is properly aligned by allocating the exact data type expected in the rust side, instead of a byte buffer in all cases.

### DIFF
--- a/rsfbclient-native/src/varchar.rs
+++ b/rsfbclient-native/src/varchar.rs
@@ -19,10 +19,12 @@ unsafe impl Send for Varchar {}
 impl Varchar {
     /// Allocate a new varchar buffer
     pub fn new(capacity: u16) -> Self {
-        let ptr = ptr::NonNull::new(unsafe {
-            alloc::alloc_zeroed(layout(capacity as usize)) as *mut InnerVarchar
+        let mut ptr = ptr::NonNull::new(unsafe {
+            alloc::alloc(layout(capacity as usize)) as *mut InnerVarchar
         })
         .unwrap();
+
+        unsafe { ptr.as_mut().len = 0 };
 
         Varchar { capacity, ptr }
     }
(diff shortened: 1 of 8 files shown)


rsfbclient-native/src/varchar.rs
```diff
@@ -0,0 +1,57 @@
...
+    ptr: ptr::NonNull<InnerVarchar>,
+}
+
+unsafe impl Send for Varchar {}
+
+impl Varchar {
+    /// Allocate a new varchar buffer
+    pub fn new(capacity: u16) -> Self {
+        let ptr = ptr::NonNull::new(unsafe {
+            alloc::alloc_zeroed(layout(capacity as usize)) as *mut InnerVarchar
```
juarezr, line 23: It's required that this allocation should be zeroed ? 

If don't, zeroing would be useless as that memory allocated here will be overwrite soon after this, Right?

Also a further improvement would be using a memory pool or memory zone for allocation for reducing the alloc/free overhead.

jairinhohw: I don't really know if the allocation needs to be zeroed, so i did it to eliminate the possibility of reading uninitialized memory from safe code, else the `as_bytes` function would need to be `unsafe`.

About the free / alloc, the `Varchar` is allocated only once per statement and reused to all rows, so i don't think it needs more optimization, but can be done in another pr maybe.

fernandobatels (approved):

rsfbclient-native/src/varchar.rs
jairinhohw: I've initialized only the `len` field to 0, so that it will be safe to call `as_bytes`.

fernandobatels (approved):

juarezr (approved):

PR #54, opened 2020-09-19
Simplify `Queryable` trait

### LINKED ISSUE #51: API Ergonomics: Consider reorganizing lifetimes in rsfbclient main crate traits/structs
The lifetime in `Queryable<'a, R>` (and `Transaction<'a,C>`, `StmtIter<'a, R,C>` etc)  will tend to infect whatever uses it.
Unfortunately, due to some warts of the rust language, this leads to unexpected compilation errors in some cases.
For example:
```
  const GTT_QRY: &str =
  "select rdb$relation_name
     from rdb$relations
    where     rdb$relation_name = 'GTT_TABLE'
          and rdb$relation_type > 3;";

  const GTT_CREATE: &str =
    "create global temporary table GTT_TABLE
     (table_id INT not null primary key)
     on commit preserve rows;";

  pub fn with_temporary_table<'a, E>( conn: &'a mut E ) -> ()
    where 
      E: Execute + Queryable<'a, (String,)>
  {
    let result : Vec<(String,)> = conn.query(GTT_QRY, ()).unwrap();
    let result2 =  conn.execute(GTT_CREATE, ());
  }

error[E0499]: cannot borrow `*conn` as mutable more than once at a time
   --> src/main.rs:100:20
    |
95  |   pub fn with_temporary_table<'a, E>( conn: &'a mut E ) -> ()
    |                               -- lifetime `'a` defined here
...
99  |     let result : Vec<(String,)> = conn.query(GTT_QRY, ()).unwrap();
    |                                   -----------------------
    |                                   |
    |                                   first mutable borrow occurs here
    |                                   argument requires that `*conn` is borrowed for `'a`
100 |     let result2 =  conn.execute(GTT_CREATE, ());
    |                    ^^^^ second mutable borrow occurs here
```
pardon if i'm misunderstanding something, but I also also don't see why R of `Queryable<'a, R>` should be bounded by `'a` at the trait level like it is in its where clause either:  `where R : FromRow + 'a`
It seems like the lifetime of R should be unrelated to the connection/transaction/iterator at the trait level: Once we have an R on hand, we should be able to do whatever we want with it, or the Connection, in either order, if our impl allows it.

jairinhohw:
Removed lifetime and trait bounds from the trait itself.

Closes #51

### DIFF
--- a/src/connection/mod.rs
+++ b/src/connection/mod.rs
@@ -446,17 +446,21 @@ where
     }
 }
 
-impl<'a, R, C> Queryable<'a, R> for Connection<C>
+impl<R, C> Queryable<R> for Connection<C>
 where
-    R: FromRow + 'a,
-    C: FirebirdClient + 'a,
+    R: FromRow + 'static,
+    C: FirebirdClient,
 {
-    type Iter = StmtIter<'a, R, C>;
+    // type Iter = StmtIter<'a, R, C>;
 
     /// Prepare, execute, return the rows and commit the sql query
     ///
     /// Use `()` for no parameters or a tuple of parameters
-    fn query_iter<P>(&'a mut self, sql: &str, params: P) -> Result<Self::Iter, FbError>
+    fn query_iter<'a, P>(
+        &'a mut self,
+        sql: &str,
+        params: P,
+    ) -> Result<Box<dyn Iterator<Item = Result<R, FbError>> + 'a>, FbError>
     where
         P: IntoParams,
     {
@@ -476,7 +480,7 @@ where
                     _marker: Default::default(),
                 };
 
-                Ok(iter)
+                Ok(Box::new(iter))
             }
             Err(e) => {
                 // Return the statement to the cache
--- a/src/connection/mod.rs
+++ b/src/connection/mod.rs
@@ -450,8 +450,6 @@ impl<C> Queryable for Connection<C>
 where
     C: FirebirdClient,
 {
-    // type Iter = StmtIter<'a, R, C>;
-
     /// Prepare, execute, return the rows and commit the sql query
     ///
     /// Use `()` for no parameters or a tuple of parameters
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -128,8 +128,6 @@ impl<'c, C> Queryable for Transaction<'c, C>
 where
     C: FirebirdClient,
 {
-    // type Iter = StmtIter<'a, R, C>;
-
     /// Prepare, execute and return the rows of the sql query
     ///
     /// Use `()` for no parameters or a tuple of parameters
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -124,17 +124,21 @@ where
     }
 }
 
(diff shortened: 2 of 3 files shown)


src/connection/mod.rs
```diff
@@ -446,19 +446,23 @@ where
...
 
-impl<'a, R, C> Queryable<'a, R> for Connection<C>
+impl<C> Queryable for Connection<C>
 where
-    R: FromRow + 'a,
-    C: FirebirdClient + 'a,
+    C: FirebirdClient,
 {
-    type Iter = StmtIter<'a, R, C>;
+    // type Iter = StmtIter<'a, R, C>;
```
fernandobatels, line 453: We can remove this commented line

src/transaction.rs
```diff
@@ -124,19 +124,23 @@ where
...
 }
 
-impl<'a, R, C> Queryable<'a, R> for Transaction<'a, C>
+impl<'c, C> Queryable for Transaction<'c, C>
 where
-    R: FromRow + 'a,
     C: FirebirdClient,
 {
-    type Iter = StmtIter<'a, R, C>;
+    // type Iter = StmtIter<'a, R, C>;
```
fernandobatels, line 131: We can remove this commented line

fernandobatels (approved):

fernandobatels:
@afsmi you can help us writing a test, in this PR, for you problem into #51 ?

src/connection/mod.rs
jairinhohw: Done

src/transaction.rs
jairinhohw: Done

fernandobatels (approved):

fernandobatels:
@afsmi you can help us writing a test, in this PR, for you problem into #51 ?

afsmi:
I saw your first request for tests a few days ago. Sorry I didn't confirm sooner.

See issue #63 - I was going to try to run some experiments and see if I could come up with a solution and include it here (like I said in a comment for the issue of this PR: I wanted to look into it more). But I never got around to that. I've been using all my `rsfbclient` time getting issues out of my brain.

My preference right now would be to consider this PR blocked, waiting for decision on #63, and to wait before writing tests.

juarezr (approved):

fernandobatels:
> @fernandobatels I thought I saw a comment from you asking me to write tests maybe a day or two earlier but I can't find it any more. Anyways, sorry I didn't respond sooner.

Don't worry. I only 're-write' my comment because i thought you didn't have received the notification.

> I've been using all my "`rsfbclient` time" unloading issues from my brain (i think I'm done with that for now).

Hehe

> See issue #63 - I was going to try to see if I could come up with a solution and include it here (like I said in a comment for the issue of this PR: I wanted to look into it more).
> My preference right now would be to consider this PR blocked, waiting for decision on #63, and to wait before writing tests.

I think the #63 deserve his own PR. We can complete this PR, and in #63 make a new decision that change this PR. This is not a problem for me :).

PR #55, opened 2020-09-21
add tests for struct Transaction

afsmi:
Before I go create some separate issues: should these tests be expected to work?

I get

``` 
Error: Sql { msg: "unsuccessful metadata update\nobject TABLE \"RSFBCLIENT_TEST_TRANSACTION\" is in use", code: -607 }
```

### DIFF
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -101,3 +101,4 @@ macro_rules! mk_tests_default {
 mod charset;
 mod params;
 mod row;
+mod transaction;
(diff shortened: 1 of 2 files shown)


jairinhohw:
> Before I go create some separate issues: should these tests be expected to work?
> 
> I get
> 
> ```
> Error: Sql { msg: "unsuccessful metadata update\nobject TABLE \"RSFBCLIENT_TEST_TRANSACTION\" is in use", code: -607 }
> ```

This error occurs when trying to delete a table while there is some transaction using it. It is a server side error, and to avoid
it, we run the tests with the `-- --test-threads 1`flag.

jairinhohw (approved):

fernandobatels (approved):

Issue #57, opened 2020-09-21
Transaction::commit consumes handle without explicitly releasing it


afsmi:
`Transaction::commit` consumes the Transaction but is actually calling `isc_commit_retaining`, where the intention is to re-use the handle.

It probably doesn't cause too many problems right now since it gets immediately dropped when it goes out of scope (which calls `isc_rollback_transaction`), but the `Drop` implementation could change so it should be explicit.

Links to source:
[main crate 1](https://github.com/fernandobatels/rsfbclient/blob/5de7786577895c886b22fce6c8c666d810cf0611/src/transaction.rs#L33-L35) 
[main crate 2](https://github.com/fernandobatels/rsfbclient/blob/5de7786577895c886b22fce6c8c666d810cf0611/src/transaction.rs#L244-L252) 
[fb-native](https://github.com/fernandobatels/rsfbclient/blob/5de7786577895c886b22fce6c8c666d810cf0611/rsfbclient-native/src/connection.rs#L219-L224)

PR #58, opened 2020-09-21
Modify Transaction::commit to use isc_commit_transaction

afsmi:
See #57

### DIFF
--- a/src/transaction.rs
+++ b/src/transaction.rs
@@ -6,6 +6,7 @@
 
 use rsfbclient_core::{FbError, FirebirdClient, FromRow, IntoParams, TrIsolationLevel, TrOp};
 use std::marker;
+use std::mem::ManuallyDrop;
 
 use super::{connection::Connection, statement::Statement};
 use crate::{connection::stmt_cache::StmtCacheData, statement::StatementData, Execute, Queryable};
@@ -31,9 +32,15 @@ where
 
     /// Commit the current transaction changes
     pub fn commit(mut self) -> Result<(), FbError> {
-        self.data.commit(self.conn)?;
-        ManuallyDrop::new(self);
-        Ok(())
+        let result = self.data.commit(self.conn);
+
+        if result.is_ok() {
+            ManuallyDrop::new(self);
+        } else {
+            let _ = self.rollback();
+        }
+
+        result
     }
 
     /// Commit the current transaction changes, but allowing to reuse the transaction
@@ -48,7 +55,9 @@ where
 
     /// Rollback the current transaction changes
     pub fn rollback(mut self) -> Result<(), FbError> {
-        self.data.rollback(self.conn)
+        let result = self.data.rollback(self.conn);
+        ManuallyDrop::new(self);
+        result
     }
 
     /// Execute the statement without returning any row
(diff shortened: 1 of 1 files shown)


src/transaction.rs
```diff
@@ -31,7 +31,7 @@ where
 
     /// Commit the current transaction changes
     pub fn commit(mut self) -> Result<(), FbError> {
-        self.data.commit_retaining(self.conn)
+        self.data.commit(self.conn)
```
jairinhohw, line 34: We can use `ManuallyDrop` ([link]) to avoid executing the `rollback` in the drop code, as that was the reason i've used
the `commit_retaining` here. What do you think?

[link]: https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html
```suggestion
        self.data.commit(self.conn)?;
        ManuallyDrop::new(self);
        Ok(())
```

fernandobatels: Good suggestion

fernandobatels (approved):
It's ok with me. But it would be nice to consider the @jairinhohw suggestion.

src/transaction.rs
afsmi: Good idea.
1. On failure we still call `drop`, right? Is it intentional to swallow a potential rollback error in this case? (like if disconnected) My understanding is that `isc_rollback_transaction` doesn't really say much on error so it's better to return the commit error if we have to choose here.
2. Is there any reason not to do something similar for `Transaction::rollback`?
3. If we also do it for `Transaction::rollback`, do you have any opinion about calling that explicitly here on failure instead of `TransactionData::rollback` (via `drop`)? Personally I would prefer `Transaction::rollback` since it's local to the impl.

jairinhohw: 1. Yes, as the `ManuallyDrop` is only called after the commit returns successfully.
2. I think it is a good idea to do it in the `rollback` too
3. I don't see any problem in calling it explicitly, and we can use the `ManuallyDrop` in the end of the failure path to avoid `rollback` being called twice.

src/transaction.rs
```diff
@@ -31,7 +32,15 @@ where
...
-        self.data.commit_retaining(self.conn)
+        let result = self.data.commit(self.conn);
+
+        if result.is_ok() {
+            ManuallyDrop::new(self);
+        } else {
+            let _ = self.rollback();
+        }
+
+        result
```
jairinhohw, line 43: If the commit fails, the transaction is not wrapped in `ManuallyDrop`, so the `rollback` is called twice.
```suggestion
        let result = self.data.commit(self.conn);

        let mut tr = ManuallyDrop::new(self);

        if result.is_err() {
            let _ = tr.data.rollback(tr.conn);
        }

        result
```
Edit: Fixed the suggestion

jairinhohw (changes requested):

src/transaction.rs
jairinhohw: Maybe it's better to just remove the explicit `rollback` and let drop do the job in case of failure

afsmi: In my change to `Transaction::commit`, the explicit rollback call on error is `Transaction::rollback` via `self.rollback()` (not `TransactionData::rollback`), so self is consumed by that `rollback()` call, and wrapped in `ManuallyDrop` there unconditionally.

jairinhohw: Oh, true, my mistake

jairinhohw (approved):

fernandobatels (approved):

PR #74, opened 2020-10-07
Refactoring backend configuration abstractions

afsmi:
See commit message for details of changes

### TODO:
  * choose better names for things and decide on aspects of interface changes in main crate
  * fix up inline documentation (ie for rustdoc)
  * fix tests broken by changes
  * check and fix exports for added/moved items to ensure all the correct things show up in the public interface
  * check/fix backend method signatures in core traits (x:T vs x: &mut T)
  * scrutinize changes to types, especially with `XYZData` types in main crate.
    After making the backend changes, I had to make many trait bound and type parameter changes here in structs and methods.
    I only changed things enough to make the typechecker stop complaining, but I'm not sure everything is correct.

### DIFF
--- a/rsfbclient-core/src/connection.rs
+++ b/rsfbclient-core/src/connection.rs
@@ -1,46 +1,49 @@
-//! Connection trait to abstract over the client implementations
+//! Trait to abstract over firebird client implementations
 
 use num_enum::TryFromPrimitive;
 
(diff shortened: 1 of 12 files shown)


afsmi:
oops. I forgot to squash things. will re-create

PR #75, opened 2020-10-07
Restructuring backend configuration abstractions

afsmi:
[direct link to diff, without CI annotations](https://github.com/afsmi/rsfbclient/compare/master...afsmi:backend_config_abstraction?diff=split)
#### core:
  * FirebirdClient split into two new traits, FirebirdClientDbOps and FirebirdClientSqlOps.

    FirebirdClientDbOps is for database/attachment management and configuration.
    It currently contains an associated type for handling attachment configuration.

    FirebirdClientSqlOps is for transaction/statement-level functionality.

    'FirebirdClient' itself becomes a wrapper trait of the other two

  * trait method 'new' removed from backend traits. It is replaced by a factory trait in the main crate as well as configuration structs for each of the default backends.

  * FirebirdClientXAttach traits are eliminated

#### native, pure_rust:
  * reworked configuration/constructors/types
  * partially deduplicated embedded/remote attachment distinction

#### main crate:
  * reworked many types and trait bounds to reflect backend changes.
    It is now pretty much just `C: FirebirdClient` everywhere in type parameters instead of specific handle types.
    Also removed many generic parameters from `XYZData` methods

  * reworked connection builders into a single generic connection configuration struct (ConnectionConfiguration) which wraps and extends the new backend configuration types.

  * added factory trait for construction of client instances.
    both the 'connect()' method of connection builders and connection pools now require an instance of this trait.

  * client-specific impls moved into new file 'connection/builder.rs'

#### native:
  * ibase macro modified. linkage selection is now trait-based
  * `NativeFbClient` gains a marker type parameter. It can be `DynLoad` or `DynLink`, which both implement `LinkageMarker`. 
    `NativeFbClient` uses an associated type in `LinkageMarker` to get a corresponding `IBase` impl.
    Note that it is possible to construct both `DynLoad` and `DynLink` regardless of feature flags, but it should not be possible to actually get a `NativeFbClient<T>` from them unless the corresponding feature is enabled.
    This "type indirection" is to hide the `IBase` types from the outside world to some extent.

#### main crate:
  * each builder now lives in its own source file (`builder_native`,`builder_pure_rust`)
  * native builder reworked. It now uses two marker types to statically track configuration status.
    this means that Invoking connect() without having chosen both a supported linkage type and a supported connection type (embedded/remote) will cause a _**compiler error.**_
  * various tweaks to traits and bounds in `connection/mod.rs`

## todo:
  * [x] fix up inline documentation (ie for rustdoc)
  * [x] fix tests broken by changes
  * [x] fix examples
  * [x] check and fix exports/feature flags for added/moved items
  *  ~check/fix backend method signatures in core traits (`x: T` vs `x: &mut T`)~ deferred
  * [x] fix pure_rust builder
  * ~consider cutting down on the number of types and traits introduced in main crate~ deferred.
  * [x] Change `README.md` to reflect builder changes
  * [x] merge changes to master since starting this PR


fernandobatels:
I think that is the right way. But for now I need see how I will 'build' a new connection(yes, I saw the todo list).

I'm trying run this code, but I don't understand the `connect()` required parameter: 
```rust
let mut conn = ConnectionConfiguration::embedded_default()
    .db_name("/tmp/examples.fdb")
    .user("SYSDBA")
    .connect()?;
```

You can fix the `embedded_select` example? With this I will catch how build a new connection.

fernandobatels:
About the `ConnectionConfiguration`:

1. I think we can keep using the `ConnectionBuilder` name. What you think?
2. We must prevent the `embedded` and `remote` parameters confusion. Today we use a dedicated implementation for `embedded` [connection builder](https://github.com/fernandobatels/rsfbclient/blob/master/src/connection/mod.rs#L233) to prevent `host`, `port` and `pass` utilization, because they are not used in the embedded connection.

afsmi:
> About the `ConnectionConfiguration`:
> 
> 1. I think we can keep using the `ConnectionBuilder` name. What you think?

I will change it back to ConnectionBuilder.
At first I was thinking about how to do 'extensible generic builder' in rust, where builder is separate from configuration struct. But I decided eventually that should be a separate issue/pr to keep changes smaller.

> 2. We must prevent the `embedded` and `remote` parameters confusion. Today we use a dedicated implementation for `embedded` [connection builder](https://github.com/fernandobatels/rsfbclient/blob/master/src/connection/mod.rs#L233) to prevent `host`, `port` and `pass` utilization, because they are not used in the embedded connection.

Maybe you know this, but, from https://www.firebirdsql.org/pdfmanual/html/ufb-cs-embedded.html, windows section:
> A Firebird embedded server DLL can also be used as a network client. That is, if a regular Firebird server is listening for connections on the target computer, you can connect to databases on that system using a network-style connect string like inca:C:\School\Databases\Pupils.fdb.

Have you ever tried this? I think this does not apply to the Linux embedded situation. I have not tried either one.

For now I will make it similar to before (prevent confusion) and we can make a decision about this another time.

fernandobatels:
> > About the `ConnectionConfiguration`:
> > 
> > 1. I think we can keep using the `ConnectionBuilder` name. What you think?
> 
> I will change it back to ConnectionBuilder.
> At first I was thinking about how to do 'extensible generic builder' in rust, where builder is separate from configuration struct. But I decided eventually that should be a separate issue/pr to keep changes smaller.

Nice

> 
> > 1. We must prevent the `embedded` and `remote` parameters confusion. Today we use a dedicated implementation for `embedded` [connection builder](https://github.com/fernandobatels/rsfbclient/blob/master/src/connection/mod.rs#L233) to prevent `host`, `port` and `pass` utilization, because they are not used in the embedded connection.
> 
> Maybe you know this, but, from https://www.firebirdsql.org/pdfmanual/html/ufb-cs-embedded.html, windows section:
> 
> > A Firebird embedded server DLL can also be used as a network client. That is, if a regular Firebird server is listening for connections on the target computer, you can connect to databases on that system using a network-style connect string like inca:C:\School\Databases\Pupils.fdb.
> 
> Have you ever tried this? I think this does not apply to the Linux embedded situation. I have not tried either one.

No, this is new to me. I only had used the embedded in Linux.

> 
> For now I will make it similar to before (prevent confusion) and we can make a decision about this another time.

Yes. We can review this after. Maybe our users never will use a embedded db in this situation(in a network).

afsmi:
> Can you fix the `embedded_select` example? With this I will get how to build a new connection.
it should compile now

I made some more changes. I've edited my opening comment

afsmi:
>Maybe you know this, but, from https://www.firebirdsql.org/pdfmanual/html/ufb-cs-embedded.html, windows section:
>>A Firebird embedded server DLL can also be used as a network client. That is, if a regular Firebird server is listening for connections on the target computer, you can connect to databases on that system using a network-style connect string like inca:C:\School\Databases\Pupils.fdb.

I did not realize but this information is **out of date for firebird 3.x**.
I only installed can configured firebird 3.0 for the first time today directly from the tar.gz linux server kit, to test embedded.

Actually `libfbembed` does not even exist for firebird 3.x releases.
There is not really such a thing as embedded-only shared lib for 3.0

fernandobatels:
> try taking away `with_dyn_xyz()` or `as_embedded()` in the example and see what happens wink

Great job

fernandobatels (approved):

fernandobatels:
I'm enjoying the changes. I liked the 'builder_native' approach.

fernandobatels:
We need to remember to update the README.md, before finish the PR

fernandobatels (approved):

fernandobatels (approved):

afsmi:
I'm not really satisfied with the configuration situation and proliferation of structs in the main crate.
I hope to deal with this in a later PR.

Part of the challenge was trying to separate responsibilities of the different sub-crates.
For example I took the view that our `Builders` really should just be convenience wrappers around client-specific config details, which should actually be provided by the client implementations. But since both `Connection` and `FirebirdConnectionManager` depended on builder this wasn't straightforward.
In rust it's a bit challenging to make such a builder in a generic way so for now the main crate still knows a lot about specific clients, but mostly it's concentrated into the builders.

I also think the `new` method with `Args` in core API was not really the responsibility of the main API.
Some clients might not want to have such a constructor (middleware clients in particular), so instead it lives in main crate now as `FirebirdClientFactory`.  I introduced it basically to satisfy the needs of `FirebirdConnectionManager` without a `Builder`
I'm unsure if this trait should exist, I think it's a bit of a dumb type really only needed by `pool` at the moment, but it has the potential to be a clean interface for hiding details of how a client configuration and a client constructor go together, taking more responsibility away from `Builder`
For me it reminds a bit of the role of 'injector' in dependency injection frameworks and maybe it could be used for more complicated things in that way.

src/connection/pool.rs
```diff
@@ -4,31 +4,42 @@
...
 {
-    pub fn new(conn_builder: ConnectionBuilder<C>) -> Self {
-        Self { conn_builder }
+    pub fn new(client_factory: F, conn_conf: FactoryAssociatedConnectionConfig<F>) -> Self {
+        Self {
+            client_factory,
+            conn_conf,
+        }
     }
 }
```
jairinhohw, line 30: I don't see a need to have the `conn_conf` as a field of the `FirebirdConnectionManager`, as it exists in all `FirebirdClientFactory` implementations.
I think it would be better to have a new function on the `FirebirdClientFactory` trait that returns a `&ConnectionConfiguration`, so it can be used in line 42 of this file.

src/transaction.rs
```diff
@@ -11,18 +11,15 @@ use std::mem::ManuallyDrop;
 use super::{connection::Connection, statement::Statement};
 use crate::{connection::stmt_cache::StmtCacheData, statement::StatementData, Execute, Queryable};
 
-pub struct Transaction<'c, C>
+pub struct Transaction<'c, C: FirebirdClient>
 where
     C: FirebirdClient,
```
jairinhohw, line 16:
```suggestion
pub struct Transaction<'c, C>
where
    C: FirebirdClient,
```

The trait bound is repeated

src/transaction.rs
```diff
@@ -117,7 +105,7 @@ where
     }
 }
 
-impl<R, C> Iterator for StmtIter<'_, R, C>
+impl<R, C: FirebirdClient> Iterator for StmtIter<'_, R, C>
 where
     R: FromRow,
     C: FirebirdClient,
```
jairinhohw, line 111:
```suggestion
impl<R, C> Iterator for StmtIter<'_, R, C>
where
    R: FromRow,
    C: FirebirdClient,
```

The trait bound is repeated

src/connection/builders/builder_native.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub fn pass<S: Into<String>>(&mut self, pass: S) -> &mut Self {
+        self.get_initialized_remote().pass = pass.into();
+        self
+    }
+}
+
+impl<A> NativeConnectionBuilder<A, ConnTypeNotConfigured> {
+    /// Configure the native client for remote connections.
+    /// This will allow configuration via the 'host', 'port' and 'pass' methods.
+    pub fn as_remote(mut self) -> NativeConnectionBuilder<A, Remote> {
```
jairinhohw, line 231: Clippy warning.
Consider changing to `into_remote` or suppressing the warning using `#[allow(clippy::wrong_self_convention)]`.

src/connection/builders/builder_native.rs
```diff
@@ -0,0 +1,301 @@
...
+    /// compile error.
+    ///
+    /// Note that the embedded builder is only tested for firebird >=3.0.
+    /// If the embedded connection fails, the client dll may attempt to use
+    /// other means of connection automatically, such as XNET or localhost.
+    ///
+    /// On firebird 3.0 and above this may be restricted via the `Providers`
+    /// config parameter of `firebird.conf` see official firebird documentation
+    /// for more information.
+    pub fn as_embedded(self) -> NativeConnectionBuilder<A, Embedded> {
```
jairinhohw, line 253: Clippy warning.
Consider changing to `into_embedded` or suppressing the warning using `#[allow(clippy::wrong_self_convention)]`.

jairinhohw (changes requested):

fernandobatels (approved):

fernandobatels (approved):

jairinhohw (approved):

PR #77, opened 2020-10-11
Support for stmt execution without cursor

fernandobatels:
Started the support for 'insert .. returing' stmt and 'execute block'.

Now it's basically WIP code

---------

TODO:

- [x] 'insert .. returning' test
- [x] 'execute block' test
- [x] update docs
- [x] execute_returnable() in transaction
    - native
    - pure
- [x] execute_returnable() in connection
    - native
    - pure

### DIFF
--- a/rsfbclient-native/src/connection.rs
+++ b/rsfbclient-native/src/connection.rs
@@ -350,7 +350,6 @@ impl FirebirdClient for NativeFbClient {
             })
             .collect::<Result<_, _>>()?;
 
-
         self.stmt_data_map.insert(handle, (xsqlda, col_buffers));
 
         Ok((stmt_type, handle))
@@ -490,7 +489,7 @@ impl FirebirdClient for NativeFbClient {
                 } else {
                     ptr::null()
                 },
-                &**out_xsqlda
+                &**out_xsqlda,
             ) != 0
             {
                 return Err(self.status.as_error(&self.ibase));
@@ -500,7 +499,8 @@ impl FirebirdClient for NativeFbClient {
         // Just to make sure the params are not dropped too soon
         drop(params);
 
-        let (_, col_buf) = self.stmt_data_map
+        let (_, col_buf) = self
+            .stmt_data_map
             .get(&stmt_handle)
             .ok_or_else(|| FbError::from("Tried to fetch a dropped statement"))?;
 
--- a/rsfbclient-native/src/connection.rs
+++ b/rsfbclient-native/src/connection.rs
@@ -350,6 +350,7 @@ impl FirebirdClient for NativeFbClient {
             })
             .collect::<Result<_, _>>()?;
 
+
         self.stmt_data_map.insert(handle, (xsqlda, col_buffers));
 
         Ok((stmt_type, handle))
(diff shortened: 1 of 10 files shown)


src/statement.rs
```diff
@@ -191,6 +191,27 @@ where
         Ok(())
     }
 
+    /// Execute the current statement with input and output paramters
```
jairinhohw, line 194: I think it is confusing to call the return `output parameters`, it's better to say that it returns a single row

jairinhohw (approved):

src/tests/row.rs
```diff
@@ -11,6 +11,85 @@ mk_tests_default! {
...
+
+        // Using exec proc
+        let (val,): (i32,) = conn.execute_returnable("execute procedure get_value", ())?;
+        assert_eq!(150, val);
+
+        Ok(())
+    }
+
+    #[test]
+    fn execute_block() -> Result<(), FbError> {
```
jairinhohw, line 48:
```suggestion
    #[test]
    #[allow(clippy::float_cmp)]
    fn execute_block() -> Result<(), FbError> {
```

Clippy is complaining

rsfbclient-native/src/connection.rs
```diff
@@ -455,4 +455,60 @@ impl FirebirdClient for NativeFbClient {
...
+            ) != 0
+            {
+                return Err(self.status.as_error(&self.ibase));
+            }
+        }
+
+        // Just to make sure the params are not dropped too soon
+        drop(params);
+
+        let (_, col_buf) = self
```
afsmi, line 502: not too important, but any reason not to just do
`(out_xsqlda, col_buf) = ...`
at the beginning?

jairinhohw: I've tried, but it gave an error for having multiple mutable borrows of self if I'm not mistaken.
However I'm working on a subsequent pr in the branch `borrowed_handles` that fixes this by not using the `statemet_data_map`.

afsmi: are you sure?
I had no problem to compile it with that change. I'm on on rustc 1.46 in case it matters.
If you feel like sharing the error, here or in private to keep noise down, I am curious

src/statement.rs
fernandobatels: Done

rsfbclient-native/src/connection.rs
jairinhohw: Nevermind, it was the one at the `rsfbclient-rust` that needed that, I've tried it in this one and it works

fernandobatels: > not too important, but any reason not to just do
> `(out_xsqlda, col_buf) = ...`
> at the beginning?

I just followed the others methods syntax.

fernandobatels: Hmm, so we need to change this now or wait for the `borrowed_handles`?

jairinhohw: I've removed this on `borrowed_handles`, so i don't think it needs to change now, but i can just merge the changes too, no problem.

fernandobatels: Ok, so we can deal with this in the `borrowed_handles`

jairinhohw:
Do you prefer for me to merge the `borrowed_handles` in this `nocursor` branch or create a new PR after this one is merged into master?

fernandobatels:
> Do you prefer for me to merge the `borrowed_handles` in this `nocursor` branch or create a new PR after this one is merged into master?

I prefer a new PR after this one

rsfbclient-rust/src/client.rs
```diff
@@ -543,6 +556,70 @@ impl FirebirdWireConnection {
...
+                // An error ocurred
+                parse_response(&mut resp)?;
+            }
+
+            if op_code != WireOp::SqlResponse as u32 {
+                return err_conn_rejected(op_code);
+            }
+
+            dbg!(&resp);
+
```
afsmi, line 606:
```suggestion

```

rsfbclient-rust/src/wire.rs
```diff
@@ -356,6 +356,33 @@ pub fn execute(tr_handle: u32, stmt_handle: u32, input_blr: &[u8], input_data: &
...
+    req.put_u32(tr_handle);
+
+    req.put_wire_bytes(input_blr);
+    req.put_u32(0); // Input message number
+    req.put_u32(if input_blr.is_empty() { 0 } else { 1 }); // Messages
+
+    req.put_slice(input_data);
+
+    req.put_wire_bytes(output_blr);
+    req.put_u32(0); // Output message number
```
afsmi, line 381: i know this is similar to with `fetch` and `execute`, but what is this doing?
Nevermind line 378.

I see it 's part of the [xdr reference previously linked to](https://www.firebirdnews.org/docs/papers/GDSDBProtocol.html#op_execute2) but what are 
`MessageNumber`, `OutPutMessageNumber` and `Messages` actually referring to?

and what is the purpose of the output blr? i see from `xsqlda_to_blr` that it's just a list of ouput types. that seems useless to the server. is it possible to do other things with it?

jairinhohw: I am not sure, but it seems you can send the input / output data in parts, so this message number is kind of an index, but i don't really know as i never seen it not be 0.

About the output blr, it has the same role as the `coerce` function in the `XSqlDa`of the native version. It tells the server what kind of data the lib is expecting, so it can coerce the data and send it in the right format. For example, we parse all integer types as 64bit numbers, so the output xsqlda tells the server to coerce any present integer type as a 64bit integer.

jairinhohw: The `Messages` seems to be something on the lines of "how many rows of data are you sending?" so we send 0 when there is no parameters and 1 when there is, but i am not 100% sure.

afsmi (approved):

PR #79, opened 2020-10-12
Simple fix for the named parameters

jairinhohw:
Simple fix to allow disabling the named parameter functionality


jairinhohw:
It may be better to find another way, but at least the high level api does not need to be changed

fernandobatels (approved):

PR #82, opened 2020-10-16
add opt-out and opt-in options to ci using labels

### LINKED ISSUE #76: CI: allow method to disable workflows for draft PRs/other situations
Sometimes on drafts we know tests won't pass, or we're just making little changes. Check annotations can be noisy and we're just spinning up a bunch of docker containers for nothing and using the free minutes (2000/month)

I can think of different ways to do this:
* using `if: github.event.pull_request.draft == false` on all jobs (this might be a bit [broken](https://github.community/t/status-of-workflows-with-no-running-jobs/16782) but there's workarounds)
* using special branches as the merge target (complication: keep them synchronized with master only)
* using labels

Some ways make it so it is still hard or impossible to accidentally get merged later without testing. these tend to be more complicated.

Other ways are just manual overrides that can be used in any situation including drafts but also regular PRs. these tend to be simpler.

@fernandobatels I think it's just a question of whether you as repository owner are ok with this or not since you are doing the merging. It's always possible to make an issue referencing changes in a fork instead of making a draft pr here.

if you're interested I can provide more details on different approaches

afsmi:
This will change the CI in a few ways.

1. doesn't run the CI on push with only workflow changes
2. checks if CI run needed when a draft PR becomes ready for review
3. checks if CI run needed when a label is added to a PR. in most cases this does not cause the CI to run but it does cause the workflow itself to run.
4. adds detection for draft status of PRs. By default the CI is skipped for draft PRs.
5. adds detection for label 'prevent-ci' on PRs (draft or not) to prevent running CI.
6. adds detection for label 'force-ci' which overrides all of the above and forces CI to be run on subsequent changes (excluding label additions)
7. force runs the CI once when the 'force-ci' label is added to a PR

note that *all* label additions trigger the workflow, but this configuration will only allow the CI to run if the added label is `force-ci`. This means if you add any label that isn't 'force-ci' it will change the ci checks to 'skipped' regardless of what they were before.
i think this is not a big deal, the old run results should still be available under 'actions' tab.
If we ever start using a lot of other labels on PRs this could start to get noisy. To change this might require some larger restructuring of the workflow

interestingly the changes seem to work within the context of this PR itself so you can test it here by adding and removing labels
and dummy edits if you want, but before merging we should be careful the about commit noise that would introduce

closes #76

### DIFF
--- a/.github/workflows/testing_changes.yml
+++ b/.github/workflows/testing_changes.yml
@@ -41,11 +41,8 @@ jobs:
       || ( (github.event.action != 'labeled') && contains( github.event.pull_request.labels.*.name, 'force-ci' ) )
       || (github.event.label.name == 'force-ci')
     steps:
-        #helpful for debugging event
-      - name: dump action context 'github'
-        env:
-          GITHUB_CONTEXT: ${{ toJson(github) }}
-        run: echo "$GITHUB_CONTEXT"
+      - name: continue
+        run: echo not skipping CI. continuing now...
 
   linting:
     name: linting
@@ -85,9 +82,9 @@ jobs:
 
   testing:
     name: testing
-    needs: linting    
     strategy:
       fail-fast: false
+      needs: linting
       matrix:
         os: ["ubuntu-20.04", "windows-latest", "macos-latest"]
         firebird: [v3, v2, v4]
--- a/.github/workflows/testing_changes.yml
+++ b/.github/workflows/testing_changes.yml
@@ -82,9 +82,9 @@ jobs:
 
   testing:
     name: testing
+    needs: linting
     strategy:
       fail-fast: false
-      needs: linting
       matrix:
         os: ["ubuntu-20.04", "windows-latest", "macos-latest"]
         firebird: [v3, v2, v4]
--- a/.github/workflows/testing_changes.yml
+++ b/.github/workflows/testing_changes.yml
@@ -85,9 +85,9 @@ jobs:
 
   testing:
     name: testing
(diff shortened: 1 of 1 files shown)


jairinhohw (approved):

.github/workflows/testing_changes.yml
```diff
@@ -12,14 +12,45 @@
...
+        && !github.event.pull_request.draft
+      )
+      || ( (github.event.action != 'labeled') && contains( github.event.pull_request.labels.*.name, 'force-ci' ) )
+      || (github.event.label.name == 'force-ci')
+    steps:
+        #helpful for debugging event
+      - name: dump action context 'github'
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "$GITHUB_CONTEXT"
```
juarezr, line 48: This `echo` can leak any restricted information or s3cre3t?

jairinhohw: You can take a look at it [here](https://github.com/fernandobatels/rsfbclient/runs/1266171580?check_suite_focus=true#step:2:589).
I did not find anything concerning, and think it cannot contain any restricted information as it is a public repository anyways.

afsmi: github is supposed to redact this information automatically you can see this in the `token: "***"` near the beginning of the output for that `echo`

but now that you mention it I can't really find enough information to be 100% sure that this will never be a problem
so I changed it

fernandobatels (approved):

PR #84, opened 2020-10-18
Using borrowed handles in the `FirebirdClient` Trait

jairinhohw:
Also removed the `stmt_data_map` `HashMap`, storing the data it the handles instead.

The public API was modified a bit by removing the `RefCell`s from the `Connection` code, as the `Statement` needs to store a mutable reference to the `Transaction` now.


fernandobatels (approved):

afsmi (commented):
@jairinhohw could you comment about the main goals of your changes, your reasoning and choices? particularly for the `FirebirdClient` API

it does not make sense to me to make all methods of `FirebirdClient` take only `&mut XYZHandle`.

if you replace all occurrences of the associated types in `FirebirdClient` as you have done, then you could likely achieve the same effect in the implementations instead of the API

```
impl FirebirdClient for Native.. {
       type DbHandle = SomeConcreteType; //where SomeConcreteType implements BorrowMut<ibase::isc_db_handle>
       ...
    }
```
or something like that.

and this way it doesn't force every implementation to have to deal with `&mut` if it doesn't want to.

I can see how because of the `XYZAttach` traits that this could be a bit troublesome.

A modification of these changes which would have some obvious merits in my view, is if the signatures were mixed:
  * `detach_database`, `drop_database`, takes just `DbHandle`, not `&mut`
  *  `free_statement` takes just `StmtHandle`, not `&mut`


Maybe part of it is getting rid of `Clone + Copy` and then having to satisfy the type checker?

jairinhohw:
> if you replace all occurrences of the associated types in `FirebirdClient` as you have done, then you could likely achieve the same effect in the implementations instead of the API
> 
> ```
> impl FirebirdClient for Native.. {
>        type DbHandle = SomeConcreteType; //where SomeConcreteType implements BorrowMut<ibase::isc_db_handle>
>        ...
>     }
> ```

Do you mean being able to get a mutable borrow using interior mutability and use only `&XYZAttach`? That could be done, but one of my goals with this change is to remove all `Cell`s and `RefCell`s, but that would require the handles to have `RefCell`s inside of them.

> A modification of these changes which would have some obvious merits in my view, is if the signatures were mixed:
> 
>     * `detach_database`, `drop_database`, takes just `DbHandle`, not `&mut`
> 
>     * `free_statement` takes just `StmtHandle`, not `&mut`

This can be changed without problem with the current implementation, i will do that

jairinhohw:
The goals where:
Use borrowed handles, instead of copied ones.
Remove the `stmt_data_map`
Remove `RefCell`s uses

It is also a preparation to make the statement cache a wrapper around the cliend, as @afsmi suggested.

fernandobatels (approved):

afsmi:
> Edit: About the last point, now that i though better, i think it will be better to see in another PR, as it will not be so easy to do because of the types that implement `Drop` in which we can only take a mutable reference to the fields.

>The goals where:
Use borrowed handles, instead of copied ones.
Remove the stmt_data_map
Remove RefCells uses
It is also a preparation to make the statement cache a wrapper around the cliend, as @afsmi suggested.

I  think we should just keep working on the PR and solve the problems rather than merge things we know we will have to undo. But if you feel strongly about it, i will click the approve button :grin: 

Since you eliminated the `RefCell`s  I have started working on some more radical changes from this PR. I hope to simplify the main crate enough so that it becomes easy to keep the core crate signatures as we want them and also start implementing other more interesting things like cache-as-middleware

You can preview here:
https://github.com/fernandobatels/rsfbclient/compare/fernandobatels:f086b76...afsmi:0687dfb
note that It doesn't compile right now, I just wanted to keep you updated. I'll probably have something ready for a PR in a couple more days.

jairinhohw:
> I think we should just keep working on the PR and solve the problems rather than merge things to master that we know we will have to undo later. But if you feel strongly about it, i will click the approve button grin Besides the core API stuff I think these goals and changes are good

I can't really think of a good way to make the functions take only `DbHandle` instead of `&mut`, other than make the handles in `Connection` and `Statement` be wrapped in `Option` to allow taking them out. Do you have any ideas @afsmi?

afsmi:
> I can't really think of a good way to make the functions take only `DbHandle` instead of `&mut`, other than make the handles in `Connection` and `Statement` be wrapped in `Option` to allow taking them out. Do you have any ideas @afsmi?

Since in this PR the borrows in `Transaction` and `Statement` were made mutable ([here](https://github.com/fernandobatels/rsfbclient/pull/84/files#diff-ca40038cc1220f9b25a1e0808df841d4b5fafcd936bfd43b9f8e39b8719a66c5R22) and [here](https://github.com/fernandobatels/rsfbclient/pull/84/files?file-filters%5B%5D=.rs#diff-fdbffa5e7a50bcc767fff75f5c23b30abacf30cebee409d37a9e8669510135ccR17))

it's pretty clear that there's no way to really use more than one simultaneous `Transaction` or `Statement` on a single connection any more (I don't know if it was even possible before. I don't remember if I ever tried.)
But if we want those while keeping our nice high level interface, we will most likely have to reintroduce some kind of inner mutability, smart pointer, or replace handles by indices into a fixed-sized set of slots or memory pool.

Personally I'm interested to try the last of those ideas, as I think it will make async easy to do in a clean way, ie one struct which is responsible for all database activity on a thread, across possibly multiple connections, and possibly communicating with other threads via channels. I think we could combine this with the `pool` module as well.
I'm also curious about experimenting with using `Statements` across multiple `Transactions` or `Connections` just to see what happens.. [apparently](http://docwiki.embarcadero.com/InterBase/2020/en/Isc_dsql_alloc_statement2()) the API allows  one to choose whether to reset statements or not on detach?

At first I will put just 1 `Transaction` and 1 `Statement` inside  `Option`, almost as you say, except all in `Connection`.
Then `Transaction` and `Statement` don't borrow anything, they just wrap each other like
`Statement( Transaction ( Connection (... FirebirdClient ) ) )` and each time you wrap, it turns a corresponding `None` into a `Some` in the `Connection`.
No more lifetime parameters.

jairinhohw:
> Since in this PR the borrows in `Transaction` and `Statement` were made mutable ([here](https://github.com/fernandobatels/rsfbclient/pull/84/files#diff-ca40038cc1220f9b25a1e0808df841d4b5fafcd936bfd43b9f8e39b8719a66c5R22) and [here](https://github.com/fernandobatels/rsfbclient/pull/84/files?file-filters%5B%5D=.rs#diff-fdbffa5e7a50bcc767fff75f5c23b30abacf30cebee409d37a9e8669510135ccR17)), it's pretty clear that there's no way to really use more than one simultaneous `Transaction` or `Statement` on a single connection any more (I don't know if it was even possible before. I don't think I tried.)
> But if we want those while keeping our nice high level interface, we will most likely have to reintroduce some kind of inner mutability, smart pointer, or replace handles by indices into a fixed-sized set of slots or memory pool.

I don't think there is a reason to allow a single connection to have more than one active `Transaction` / `Statement` at a time, but to allow this we would need to use `Cells` and `RefCells` to change the handles from mutable references to just references.
 
> Personally I'm interested to try the last of those ideas, as I think it will make async easy to do in a clean way, ie one struct which is responsible for all database activity on a thread, across possibly multiple connections, and possibly communicating with other threads via channels. I think we could combine this with the `pool` module as well.

I don't think there is much reason to make an async version, as the `libfbclient` functions are blocking.

> I'm also curious about experimenting with using `Statements` across multiple `Transactions` or `Connections` just to see what happens.. [apparently](http://docwiki.embarcadero.com/InterBase/2020/en/Isc_dsql_alloc_statement2()) the API allows one to choose whether to reset statements or not on detach?

Statements can (and are being, because of the statement cache) used in different Transactions, but cannot be used between Connections (All handles are exclusive to their connections).
About the resetting of the `isc_dsql_alloc_statement2` i think it is just an implementation difference, not a different behavior of the allocated Statements. As a note, we cannot use this function, as the code could try to set the handles to `null` but the handles are on Rust side and are moved from the original position in memory, that would cause invalid writes (The way we are currently storing the handles).

> At first I will put just 1 `Transaction` and 1 `Statement` inside `Option`, almost as you say, except all in `Connection`.
> Then `Transaction` and `Statement` don't borrow anything, they just wrap each other like
> `Statement( Transaction ( Connection (... FirebirdClient ) ) )` and each time you wrap, it turns a corresponding `None` into a `Some` in the `Connection`.
> No more lifetime parameters.

That could be done, however crating a `Transaction`  would consume the `Connection`, so it could not be used in the `pool`, as you can only get a mutable reference to the `Connection` (As it needs to go back to the pool).

afsmi:
@jairinhohw Thanks for your insights. I see your points.

as it turns out there's also another reason why the newtype wrappers don't really make sense: creating a transaction/statement can fail, and returning `Result<... , (Self, FbError)>` is ugly.

I find the situation in the main crate a sort of unsatisfactory, I will see if I can come up with some alternative. I do feel that I've made some progress in slimming down the main crate.

afsmi (approved):

PR #86, opened 2020-10-25
Diesel

fernandobatels:
Todo:

- [x] insert support
- [x] delete support
- [x] update support
- [x] fix CI erros
- [x] simple select support
- [x] select first support
- [x] filter support
- [x] order by support
- [x] limit support
- [x] limit offset support
- [x] find support
- [x] distinct support
- params support:
    - [x] text/varchar
    - [x] int
    - [x] float
    - [x] blob
    - [x] date/time
    - [x] bool
    - [x] null
- columns result support:
    - [x] text/varchar
    - [x] int
    - [x] float
    - [x] blob
    - [x] date/time
    - [x] bool
    - [x] null
- variations:
  - [x] native dynamic
  - [x] native linking
  - [x] native dynamic embedded
  - [x] native linking embedded
  - [x] pure rust 
- [x] transactions
- [x] string connection
- [x] Use something like the `SupportsDefaultKeyword` instead the 'no-default-stmt-impl' feature


rsfbclient-diesel/src/fb/transaction.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn default() -> Self {
+        FbTransactionManager::new()
+    }
+}
+
+impl<'c> TransactionManager<FbConnection<'c>> for FbTransactionManager<'c> {
+    fn begin_transaction(&self, conn: &FbConnection) -> QueryResult<()> {
+        let depth = self.depth.get() + 1;
+        if depth == 1 {
+            let conn = unsafe { conn.raw.as_ptr().as_mut().unwrap() };
```
jairinhohw, line 36: I've created a new tests called `transaction_depth_multithreaded` which causes random crashes, probably because of this unsafe block, but i still am not sure what exactly is happening.
I'll try to find a way to remove the unsafe here to see if the problem is fixed

jairinhohw (changes requested):

rsfbclient-diesel/src/fb/transaction.rs
fernandobatels: I'm trying resolve this bug too. My initial idea is use the [borrow_mut](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow_mut), but this results in the [E0495 error](https://doc.rust-lang.org/error-index.html#E0495)

fernandobatels: Well, I believe the problem was not exactly in the unsafe block. I changed the `transaction_depth_multithreaded` test to use the `sql_query` instead the default query builder.

Unfortunately, now tarpaulin that is showing an error. I will check this.

rsfbclient-diesel/src/fb/connection.rs
```diff
@@ -0,0 +1,194 @@
...
+use diesel::result::Error::DeserializationError;
+use diesel::result::*;
+use rsfbclient::SimpleConnection as FbRawConnection;
+use rsfbclient::{Execute, Queryable, SqlType};
+use std::cell::RefCell;
+
+pub struct FbConnection<'c> {
+    pub raw: RefCell<FbRawConnection>,
+    tr_manager: FbTransactionManager<'c>,
+}
```
jairinhohw, line 22: The main problem seems to be here, as the `tr_manager` needs to hold a reference to the `raw`, making it a self referential struct, so this `<'c>` lifetime does not make sense, making that unsafe block in the `begin_transaction` create an potentially invalid lifetime (because it creates a lifetime that matches `'c` with unsafe code)

jairinhohw:
@fernandobatels please take a look at my changes.
I've created a default transaction on the `Connection` type, so that there is no need to create and store a `Transaction` in the
`TransactionManager`.
Also changed the behavior of the `Queryable` and `Executable` methods in case the new `begin_transaction` method on the `Connection` was called and there is an active transaction.

fernandobatels:
> @fernandobatels please take a look at my changes.
> I've created a default transaction on the `Connection` type, so that there is no need to create and store a `Transaction` in the
> `TransactionManager`.
> Also changed the behavior of the `Queryable` and `Executable` methods in case the new `begin_transaction` method on the `Connection` was called and there is an active transaction.

I liked the changes. Previously I thought about creating a `begin_transaction` that would return a new transaction, but the idea of a default transaction is very nice.

My only concern is with the `use_transaction` method. I'm afraid that this may cause some confusion with the `with_transaction` method. Is it necessary to be public?

> Edit:. Now it only fails in the coverage, but this is probably some problem of the `tarpaulin` with threaded code

I'm running the CI again, to see if this persists

jairinhohw:
> My only concern is with the `use_transaction` method. I'm afraid that this may cause some confusion with the `with_transaction` method. Is it necessary to be public?

It can be private, as my initial idea was to use it directly in the `rsfbclient-diesel` crate, but then i created the `begin_transaction`, `commit` and `rollback` functions and forgot to remove the `pub` on the `use_transaction`.

I will try to find a better test that fails in the older version without using threaded code to see if tarpaulin stops complaining

jairinhohw:
I've replaced the multithreaded test with a simpler version that causes the same segfault on the version with unsafe code, hopefully tarpaulin will not complain anymore

.github/workflows/testing_changes.yml
```diff
@@ -221,28 +224,84 @@ jobs:
...
+        with:
+          command: test
+          args: -p rsfbclient-rust
+
+      - name: Run cargo test on rsfbclient-diesel
+        uses: actions-rs/cargo@v1
+        if: matrix.plataform != 'macos' || matrix.build != 'linking'
+        with:
+          command: test
+          args: --manifest-path rsfbclient-diesel/Cargo.toml --no-default-features --features '${{ matrix.features_diesel }}'' -- --test-threads 1
```
dancingbug, line 246:
```suggestion
          args: --manifest-path rsfbclient-diesel/Cargo.toml --no-default-features --features '${{ matrix.features_diesel }}' -- --test-threads 1
```

btw string literals in workflows are supposed to be single quotes
https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#literals

but if it's working fine right now i'm ok not to touch the rest of the file

rsfbclient-diesel/src/fb/transaction.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn commit_transaction(&self, conn: &FbConnection) -> QueryResult<()> {
+        conn.raw
+            .borrow_mut()
+            .commit()
+            .map_err(|e| DatabaseError(DatabaseErrorKind::__Unknown, Box::new(e.to_string())))?;
+
+        self.depth.set(0);
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
fernandobatels: You are right. Fixed.

.github/workflows/testing_changes.yml
fernandobatels: Fixed

jairinhohw (approved):

fernandobatels:
I rebase the branch, to improve the history commits

fernandobatels:
Note: When this PR is approved, I will open a PR for the `diesel` repo with [my changes](https://github.com/diesel-rs/diesel/compare/master...fernandobatels:master).

rsfbclient-diesel/src/fb/connection.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl Connection for FbConnection {
+    type TransactionManager = FbTransactionManager;
+    type Backend = Fb;
+
+    fn establish(database_url: &str) -> ConnectionResult<Self> {
+        #[cfg(feature = "pure_rust")]
+        let mut raw_builder = rsfbclient::builder_pure_rust();
+        #[cfg(any(feature = "linking", feature = "dynamic_loading"))]
```
dancingbug, line 41:
```suggestion
        // TODO: generalize to other backends
```
can we remove this at least for now?

1. I think this is not a good way to do it (not modular, and if `--all-features` is given then it does something unexpected)
2. `FbRawConnection` is hardcoded to native anyways

rsfbclient-native/src/ibase.rs
```diff
@@ -650,9 +650,9 @@ parse_functions! {
     //         arg9: *mut ISC_SCHAR,
     //     ) -> ISC_STATUS;
     // }
-    // extern "C" {
-    //     pub fn isc_vax_integer(arg1: *const ISC_SCHAR, arg2: ::std::os::raw::c_short) -> ISC_LONG;
-    // }
+    extern "C" {
+        pub fn isc_vax_integer(arg1: *const ISC_SCHAR, arg2: ::std::os::raw::c_short) -> ISC_LONG;
+    }
```
dancingbug, line 655:
```suggestion
//    extern "C" {
//        pub fn isc_vax_integer(arg1: *const ISC_SCHAR, arg2: ::std::os::raw::c_short) -> ISC_LONG;
//    }
```
1. does not seem to be used anywhere
2. i think it's deprecated in favor of `isc_portable_integer` ([link](https://firebird-devel.narkive.com/cmZiIKZI/extending-precision-of-api-information-values)). this is definitely the case for interbase ([link](http://docwiki.embarcadero.com/InterBase/2020/en/Isc_vax_integer()))

src/statement.rs
```diff
@@ -32,10 +32,11 @@ where
         Ok(Statement { data, tr })
     }
 
-    /// Execute the current statement without returnig any row
+    /// Execute the current statement without returnig the
+    /// affecteds rows
```
dancingbug, line 36:
```suggestion
    /// Execute the current statement, returning a
    /// count of affected rows
```

rsfbclient-rust/src/client.rs
```diff
@@ -455,7 +459,18 @@ impl FirebirdWireConnection {
...
+        // Read lazy responses
+        for _ in 0..self.lazy_count {
+            if op_code != WireOp::Response as u32 {
+                return err_conn_rejected(op_code);
+            }
+            self.lazy_count -= 1;
+            parse_response(&mut resp)?;
+
+            op_code = resp.get_u32()?;
+        }
```
dancingbug, line 473: @jairinhohw I think you probably authored this-

Could you elaborate a bit on the lazy response reading?

some specific questions:
1. Is the goal to help avoid syscalls?
2. Maybe a naive question, which applies not only here- does the ordering of response XDR structures by fb server necessarily match TCP ordering of API calls? Do docs say somewhere that this is so? 

any additional info is also appreciated

dancingbug:
@jairinhohw or @fernandobatels could one of you open a new PR to merge https://github.com/fernandobatels/rsfbclient/tree/diesel_split with master, and add me as reviewer? (If I create it, I can't assign myself)

It's a subset of this branch corresponding to all the non-diesel changes

PR #91, opened 2020-11-23
Connection string support

fernandobatels:
- [x] Update the readme with the new way of make a connection
- Native impl
  - [x] Remote && Linked
  - [x] Embedded && Linked
  - [x] Remote && Dyn
  - [x] Embedded  && Dyn
- [x] Pure rust impl
- Parameters:
  - [x] host
  - [x] user
  - [x] pass
  - [x] port
  - [x] db path
  - [x] charset
  - [x] stmt size
  - [x] dialect
  - [x] fbclient path

### DIFF
--- a/rsfbclient-core/src/connection.rs
+++ b/rsfbclient-core/src/connection.rs
@@ -139,10 +139,10 @@ impl FromStr for Dialect {
     type Err = FbError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.parse::<u8>() {
-            Ok(1) => Ok(Dialect::D1),
-            Ok(2) => Ok(Dialect::D2),
-            Ok(3) => Ok(Dialect::D3),
+        match s {
+            "1" => Ok(Dialect::D1),
+            "2" => Ok(Dialect::D2),
+            "3" => Ok(Dialect::D3),
             _ => Err(FbError::from(format!(
                 "'{}' doesn't represent any dialect",
                 s
--- a/rsfbclient-core/src/connection.rs
+++ b/rsfbclient-core/src/connection.rs
@@ -1,6 +1,7 @@
 //! Traits to abstract over firebird client implementations
 
 use num_enum::TryFromPrimitive;
+use std::str::FromStr;
 
 use crate::*;
 
@@ -132,6 +133,22 @@ pub enum Dialect {
     D1 = 1,
     D2 = 2,
     D3 = 3,
+}
+
+impl FromStr for Dialect {
+    type Err = FbError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.parse::<u8>() {
+            Ok(1) => Ok(Dialect::D1),
+            Ok(2) => Ok(Dialect::D2),
+            Ok(3) => Ok(Dialect::D3),
+            _ => Err(FbError::from(format!(
+                "'{}' doesn't represent any dialect",
+                s
+            ))),
+        }
+    }
 }
 
 #[repr(u8)]
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -328,7 +328,7 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
     /// - If you not provide the host, we will consider this a embedded connection.
     /// - The port, user and pass parameters only will be accepted if you provide the host.
     /// - If you not provide the `lib={fbclient}`, we will consider this a dynamic linked connection. The default, by the way.
-    pub fn with_string(
+    pub fn from_string(
         self,
         s_conn: &str,
     ) -> Result<NativeConnectionBuilder<DynByString, ConnByString>, FbError> {
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -328,6 +328,7 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
     /// - If you not provide the host, we will consider this a embedded connection.
     /// - The port, user and pass parameters only will be accepted if you provide the host.
     /// - If you not provide the `lib={fbclient}`, we will consider this a dynamic linked connection. The default, by the way.
+    #[allow(clippy::wrong_self_convention)]
     pub fn from_string(
         self,
         s_conn: &str,
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -318,6 +318,7 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
     ) -> Result<NativeConnectionBuilder<L, Remote>, FbError> {
         let settings = conn_string::parse(s_conn)?;
 
+
         let mut cb = {
             // When we have a host, we will consider
             // this a remote connection
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -374,6 +374,10 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
             cb.lib_path = Some(lib_path);
         }
 
+        if let Some(stmt_cache_size) = settings.stmt_cache_size {
+            cb.stmt_cache_size(stmt_cache_size);
+        }
+
         Ok(cb.safe_transmute())
     }
 }
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -328,9 +328,9 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
     /// - If you not provide the host, we will consider this a embedded connection.
     /// - The port, user and pass parameters only will be accepted if you provide the host.
     /// - If you not provide the `lib={fbclient}`, we will consider this a dynamic linked connection. The default, by the way.
-    pub fn with_string<S: Into<String>>(
+    pub fn with_string(
         self,
-        s_conn: S,
+        s_conn: &str,
     ) -> Result<NativeConnectionBuilder<DynByString, ConnByString>, FbError> {
         let settings = conn_string::parse(s_conn)?;
 
--- a/src/connection/builders/builder_native.rs
+++ b/src/connection/builders/builder_native.rs
@@ -1,4 +1,5 @@
 use super::*;
+use crate::connection::conn_string;
 use std::marker::PhantomData;
 
 #[doc(hidden)]
@@ -157,7 +158,7 @@ where
         self
     }
 
-    /// SQL Dialect. Default: 3
+    /// Connection charset. Default: UTF-8
     pub fn charset(&mut self, charset: Charset) -> &mut Self {
         self.charset = charset;
         self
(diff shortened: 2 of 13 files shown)


jairinhohw:
The [`url`](https://docs.rs/url/2.2.0/url/index.html) crate can be used to parse the connection string, instead of doing it manually.

fernandobatels:
> The [`url`](https://docs.rs/url/2.2.0/url/index.html) crate can be used to parse the connection string, instead of doing it manually.

I will check

fernandobatels:
> The [`url`](https://docs.rs/url/2.2.0/url/index.html) crate can be used to parse the connection string, instead of doing it manually.

Now I have all the params working. I don't think the crate URL is necessary. So for now, I will keep the original implementation.

src/connection/conn_string.rs
```diff
@@ -0,0 +1,432 @@
...
+        pass,
+        host,
+        port,
+        db_name,
+        charset,
+        dialect,
+        lib_path,
+        stmt_cache_size,
+    })
+}
```
afsmi, line 119: @fernandobatels

I have extracted each regex into a `const &str` here: [link](https://github.com/afsmi/rsfbclient/blob/connstr_regex_tweaks/src/connection/conn_string.rs#L31-L40)

to save you a click I will copy it here:
```
    const RGX_USER: &str = r#"(?:(/))([[:alnum:]]+)(?:.*)(?:@)"#;
    const RGX_PASS: &str = r#"(?:(:))([[:alnum:]]+)(?:(@))"#;
    const RGX_HOST: &str = r#"((?:://)|(?:@))([^@/:]+)((?:\w:/)|(?::[[:digit:]])|(?:/))"#;
    const RGX_PORT: &str = r#"(?:(:))([[:digit:]]+)(?:(/))"#;
    const RGX_DB_NAME: &str = r#"((?:@\w+/)|(?:[0-9]/))([^\?]+)"#;
    const RGX_DB_NAME_EMBEDDED: &str = r#"(?://)([^\?]+)"#;
    const RGX_LIBPATH: &str = r#"(?:\?)(?:.*)(lib=)([^&]+)"#;
    const RGX_DIALECT: &str = r#"(?:\?)(?:.*)(dialect=)([[:digit:]])"#;
    const RGX_CHARSET: &str = r#"(?:\?)(?:.*)(charset=)([^&]+)"#;
    const RGX_STMT_CACHE_SIZE: &str = r#"(?:\?)(?:.*)(stmt_cache_size=)([[:digit:]]+)"#;
```



Could you provide some explanations (either via adding inline code comments, or here in the PR ):
* intended behavior of each regex, including:
  * the purpose of non-obvious capture groups (use your judgment), for example:
    * in `RGX_HOST`: why is there a `://` group at the beginning? is it to handle hosts of the form `uri_scheme://...` ?
    * in `RGX_DB_NAME`: what is the intention of `(?:@\w+/)|(?:[0-9]/))` at the beginning? I think it is to start matching only after the host, but the intention isn't really clear
  * a list of format variations that are intended to be handled
* which combinations of sometimes-optional parameters should be allowed and disallowed at the parser level (ignoring any validation that may be done later).

also: i believe usernames and passwords are allowed to contain special characters including `@` and `:`, though in the case of using `isql` to connect, they are then supposed to be double quoted. do you want to support that?
[source](https://www.firebirdsql.org/file/documentation/html/en/firebirddocs/qsg3/firebird-3-quickstartguide.html#qsg3-databases-connecting-isql)  (do a text search for "`funny stuff`")

If so, in addition to being careful about what is actually being matched, you will need some kind of escape or extra delimiters to disambiguate parses. For instance
```
parse(r#"firebird://a@::@@192.168.0.1//srv/db/database_name.fdb?dialect=3"#)
```
in this example, is (user,pass) meant to be `(a@:,@)` or `(a@,:@)` ?
(also i am not sure if double quotes themselves are supposed to be allowed, then there are additional edge cases if we support double quotes in usernames/passwords and also use them as delimiters)

if you prefer to just not support usernames/passwords containing `@` or `:`, `"` or other symbols for now, I'd consider that acceptable, so long as there is a doc comment about it.

rsfbclient-core/src/connection.rs
```diff
@@ -134,6 +135,22 @@ pub enum Dialect {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.parse::<u8>() {
+            Ok(1) => Ok(Dialect::D1),
+            Ok(2) => Ok(Dialect::D2),
+            Ok(3) => Ok(Dialect::D3),
+            _ => Err(FbError::from(format!(
+                "'{}' doesn't represent any dialect",
+                s
+            ))),
+        }
```
jairinhohw, line 150: Just to simplify a bit.

```suggestion
        match s {
            "1" => Ok(Dialect::D1),
            "2" => Ok(Dialect::D2),
            "3" => Ok(Dialect::D3),
            _ => Err(FbError::from(format!(
                "'{}' doesn't represent any dialect",
                s
            ))),
        }
```

src/connection/builders/builder_native.rs
```diff
@@ -306,4 +318,66 @@ impl<A> NativeConnectionBuilder<LinkageNotConfigured, A> {
...
+    /// Basic string format: `firebird://{user}:{pass}@{host}:{port}/{db_name}?charset={charset}&lib={fbclient}&dialect={dialect}`
+    ///
+    /// Some considerations:
+    /// - If you not provide the host, we will consider this a embedded connection.
+    /// - The port, user and pass parameters only will be accepted if you provide the host.
+    /// - If you not provide the `lib={fbclient}`, we will consider this a dynamic linked connection. The default, by the way.
+    pub fn with_string<S: Into<String>>(
+        self,
+        s_conn: S,
+    ) -> Result<NativeConnectionBuilder<DynByString, ConnByString>, FbError> {
```
jairinhohw, line 334: There is no need to own the string.

```suggestion
    pub fn with_string(
        self,
        s_conn: &str,
    ) -> Result<NativeConnectionBuilder<DynByString, ConnByString>, FbError> {
```

src/connection/builders/builder_pure_rust.rs
```diff
@@ -80,11 +81,54 @@ impl PureRustConnectionBuilder {
...
     }
+
+    /// Setup the connection using the string
+    /// pattern.
+    ///
+    /// You can use the others methods(`host()`,`user()`...) to config
+    /// some default values.
+    ///
+    /// Basic string format: `firebird://{user}:{pass}@{host}:{port}/{db_name}?charset={charset}&dialect={dialect}`
+    pub fn with_string<S: Into<String>>(&mut self, s_conn: S) -> Result<&mut Self, FbError> {
```
jairinhohw, line 97: No need to own the string.

```suggestion
    pub fn with_string(&mut self, s_conn: &str) -> Result<&mut Self, FbError> {
```

src/connection/conn_string.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub dialect: Option<Dialect>,
+    pub lib_path: Option<String>,
+    pub stmt_cache_size: Option<usize>,
+}
+
+/// Parse the connection string.
+///
+/// Basic string sintax: `firebird://{user}:{pass}@{host}:{port}/{db_name}?{options}`
+pub fn parse<S: Into<String>>(conn_s: S) -> Result<ConnStringSettings, FbError> {
+    let sconn = conn_s.into();
```
jairinhohw, line 23: There is no need to own the string.

```suggestion
pub fn parse(sconn: &str) -> Result<ConnStringSettings, FbError> {
```

src/connection/builders/builder_native.rs
jairinhohw: Also, i think the name `from_string` would be better, as we are constructing a connection from a connection string.

jairinhohw (changes requested):

src/connection/conn_string.rs
jairinhohw: I think it is better to use the [`url`](https://docs.rs/url/2.2.0/url/index.html) crate to avoid parsing problems such as those ones.

fernandobatels: Ok, I didn't think about the passwords with '@'. I will try use the url crate to avoid this problems instead of try explain all the regex

jairinhohw (approved):

src/connection/conn_string.rs
```diff
@@ -0,0 +1,430 @@
...
+    pub db_name: String,
+    pub charset: Option<Charset>,
+    pub dialect: Option<Dialect>,
+    pub lib_path: Option<String>,
+    pub stmt_cache_size: Option<usize>,
+}
+
+/// Parse the connection string.
+///
+/// Basic string sintax: `firebird://{user}:{pass}@{host}:{port}/{db_name}?{options}`
```
juarezr, line 21: s/sintax/syntax

src/connection/conn_string.rs
```diff
@@ -0,0 +1,430 @@
...
+pub fn parse(sconn: &str) -> Result<ConnStringSettings, FbError> {
```
juarezr, line 22: Some questions:
- It follows IETF RFC [3986](https://tools.ietf.org/html/rfc3986) ? (Eg.: RFC [7869](https://tools.ietf.org/html/rfc7869))
- It's worth any provision for handling distinct drivers/implementations/modes like [SQLAlchemy](https://docs.sqlalchemy.org/en/14/dialects/mysql.html#mariadb-support) does?

afsmi:
questions:
1. how do I use special characters in the username and password? should `parse` translate `%40` into `@`, for example? Will users be expected to %-encode special characters themselves?
2. if i'm understanding the doc and code comments correctly (as well as from my testing of certain edge cases), is it right to say there are only two formats that are explicitly supported, and anything else is just coincidental?. The supported formats would be:
    * ```firebird://<filepath | db_alias>[?<opt>=<val>[&<opt>=<val> ...]]```
    * ```firebird://<user>[:[<pass>]]@<host>[:[<port>]]/<filepath | db_alias>[?<opt>=<val>[&<opt>=<val> ...]]```
   
    in particular,
      *  usernames are not supported for embedded, ( i think it only effectively matters if one is using a mixture of connection methods and multiple usernames/passwords. see [here](https://stackoverflow.com/a/63609951/1212536) )
      * more specifically, the user must either provide both of `<user>`, `<host>`, or neither of them.
      * if `<user>` and `<host>` are provided, each of `<pass>` and `<port>` may be blank. When they are blank, the `:` that precedes them may still be present

    is that correct?

src/connection/conn_string.rs
jairinhohw: >     * It follows IETF RFC [3986](https://tools.ietf.org/html/rfc3986) ? (Eg.: RFC [7869](https://tools.ietf.org/html/rfc7869))
 The crate used for parsing follows [this](https://url.spec.whatwg.org/) spec. It seems to be similar, but not identical to RFC 3986 and RFC 3987.

jairinhohw:
>     1. how do I use special characters in the username and password? should `parse` translate `%40` into `@`, for example? Will users be expected to %-encode special characters themselves?
By doing some testing, using `%40` (or any other `%xx` variation) works, as well as just inputting the special character directly in most cases (did not find a case it fails yet), however, special characters are returned from the parser in the `%xx` format, so we need to decode with the [`percent_encoding`](https://docs.rs/percent-encoding/2.1.0/percent_encoding/) crate.

src/connection/conn_string.rs
fernandobatels: > It follows IETF RFC [3986](https://tools.ietf.org/html/rfc3986) ? (Eg.: RFC [7869](https://tools.ietf.org/html/rfc7869))

Yes, I think. Others databases use this same 'string layout'. Why?

fernandobatels: > It's worth any provision for handling distinct drivers/implementations/modes like [SQLAlchemy](https://docs.sqlalchemy.org/en/14/dialects/mysql.html#mariadb-support) does?

Sorry, I'm not following you. To choose between our two variations(pure rust/native) you must use the respective connection builder, like `builder_native()`. And to use the dynamic linked or dynamic loaded, you use the `?lib=` parameter.

fernandobatels:
>     1. how do I use special characters in the username and password? should `parse` translate `%40` into `@`, for example? Will users be expected to %-encode special characters themselves?

I will test, but I think @jairinhohw is right.

fernandobatels:
>     2\. if i'm understanding the doc and code comments correctly (as well as from my testing of certain edge cases), is it right to say there are only two formats that are explicitly supported, and anything else is just coincidental?. The supported formats would be:

Yes. We basically support only these two variations.

> usernames are not supported for embedded, 

Yes. We already don't support set the user in the embedded connection builder.

> more specifically, the user must either provide both of <user>, <host>, or neither of them.

Yes

> if <user> and <host> are provided, each of <pass> and <port> may be blank. When they are blank, the : that precedes their usual position may still be present

Yes

src/connection/conn_string.rs
fernandobatels: Fixed.

fernandobatels:
> questions:
> 
>     1. how do I use special characters in the username and password? should `parse` translate `%40` into `@`, for example? Will users be expected to %-encode special characters themselves?

Added the support for special chars. 

> if <user> and <host> are provided, each of <pass> and <port> may be blank. When they are blank, the : that precedes their usual position may still be present

Added tests for this cases.

afsmi (approved):

jairinhohw (approved):

PR #93, opened 2020-12-10
SimpleConnection created

fernandobatels:
A new API, called SimpleConnection, created. Intended for when we need work with multiple connection types dynamically.

- [x] SimpleConnection
- [x] SimpleTransaction
- [x] Update docs
- [x] Update README.md

----------------

My initial idea was move our current `Connection<c>` and `Transaction<c>` types to an 'inner' struct, but this will demand lot of work and change the code a lot. 

So, to simplify, I made two new types: `SimpleConnection` and `SimpleTransaction`. This two type encapsulate the original structs, but works like a optional approach. To use, you just need call `.into()` method.

This change will help me in the [diesel PR](#86).


src/transaction/simple.rs
```diff
@@ -0,0 +1,271 @@
...
+#[cfg(feature = "linking")]
+impl<'c> From<Transaction<'c, NativeFbClient<DynLink>>> for SimpleTransaction<'c> {
+    fn from(tr: Transaction<'c, NativeFbClient<DynLink>>) -> Self {
+        let inner = TypeTransactionContainer::NativeDynLink(tr);
+        SimpleTransaction { inner }
+    }
+}
+
+#[cfg(feature = "linking")]
+impl<'c> From<SimpleTransaction<'c>> for Result<Transaction<'c, NativeFbClient<DynLink>>, FbError> {
```
dancingbug, line 41: why not `TryFrom`?

src/transaction/simple.rs
```diff
@@ -0,0 +1,271 @@
...
+            #[cfg(feature = "linking")]
+            TypeTransactionContainer::NativeDynLink(tr) => tr.execute_immediate(sql),
+            #[cfg(feature = "dynamic_loading")]
+            TypeTransactionContainer::NativeDynLoad(tr) => tr.execute_immediate(sql),
+            #[cfg(feature = "pure_rust")]
+            TypeTransactionContainer::PureRust(tr) => tr.execute_immediate(sql),
+        }
+    }
+
+    // TODO: add the prepare() method
```
dancingbug, line 165: not a blocker, since the old functionality is still present in `transaction/mod.rs`, but-
 is there some special challenge for implementing `prepare()` for `SimpleTransaction`?

seems like it would be "*simple*"

jairinhohw (commented):
Nothing to note, except what @dancingbug has commented already

src/transaction/simple.rs
fernandobatels: No, it's not simple. To provide the `prepare` method we need implement a ´StatementSimple´ too. I think this is not a problem for now and if someone ask this, we can review and find another approach.

src/transaction/simple.rs
fernandobatels: Because I forgot about `TryFrom` hehe. I will fix.

dancingbug (approved):

juarezr (approved):
Nice to see working towards integrating with Diesel.

jairinhohw (approved):

PR #94, opened 2020-12-23
Utils API

fernandobatels:
A new API with some methods to help the access on the [firebird context variables](https://github.com/FirebirdSQL/firebird/blob/master/doc/sql.extensions/README.context_variables2).

- db_name
   - [x] SimpleConn
   - [x] Conn
- engine_version
   - [x] SimpleConn
   - [x] Conn
   - [x] impl Cmp
   - [x] impl Ord

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -2,14 +2,12 @@
 //! Some API utils
 //!
 
-use rsfbclient_core::{FirebirdClient};
-use crate::{FbError, Connection, SimpleConnection, Queryable};
+use crate::{FbError, Queryable};
 use std::cmp::*;
 use std::convert::From;
 
 /// Infos about the server, database, engine...
 pub trait SystemInfos {
-
     /// Return the current connected database name
     fn db_name(&mut self) -> Result<String, FbError>;
 
@@ -31,7 +29,7 @@ impl From<EngineVersion> for u8 {
             EngineVersion::V1 => 1,
             EngineVersion::V2 => 2,
             EngineVersion::V3 => 3,
-            EngineVersion::V4 => 4
+            EngineVersion::V4 => 4,
         }
     }
 }
@@ -50,13 +48,17 @@ impl PartialOrd for EngineVersion {
 
 impl Eq for EngineVersion {}
 
-impl<C: FirebirdClient> SystemInfos for Connection<C> {
-
+impl<T> SystemInfos for T
+where
+    T: Queryable,
+{
     fn db_name(&mut self) -> Result<String, FbError> {
-        let (name,): (String,) = self.query_first(
-            "SELECT rdb$get_context('SYSTEM', 'DB_NAME') from rdb$database;",
-            (),
-        )?.unwrap();
+        let (name,): (String,) = self
+            .query_first(
+                "SELECT rdb$get_context('SYSTEM', 'DB_NAME') from rdb$database;",
+                (),
+            )?
+            .unwrap();
 
         Ok(name)
     }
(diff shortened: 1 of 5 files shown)


src/utils.rs
```diff
@@ -42,6 +42,12 @@ impl PartialEq for EngineVersion {
     }
 }
 
+impl PartialOrd for EngineVersion {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some((*self as u8).cmp(&(*other as u8)))
+    }
+}
```
dancingbug, line 47: you can just do
`#[derive(PartialOrd, PartialEq, Eq, Debug, Copy, Clone)]`
on `EngineVersion`

src/utils.rs
```diff
@@ -56,17 +62,48 @@ impl<C: FirebirdClient> SystemInfos for Connection<C> {
...
+            "SELECT rdb$get_context('SYSTEM', 'ENGINE_VERSION') from rdb$database;",
+            (),
+        )?;
+
+        if let Some((version,)) = row {
+            return match version {
+                ver if ver.starts_with("4.") => Ok(EngineVersion::V4),
+                ver if ver.starts_with("3.") => Ok(EngineVersion::V3),
+                ver if ver.starts_with("2.") => Ok(EngineVersion::V2),
+                ver => Err(FbError::from(format!("Version not detected: {}", ver)))
```
dancingbug: little bit shorter.

```suggestion
            return match &version[0..2] {
                "4." => Ok(EngineVersion::V4),
                "3." => Ok(EngineVersion::V3),
                "2." => Ok(EngineVersion::V2),
                _ => Err(FbError::from(format!("Version not detected: {}", version)))
```

src/utils.rs
```diff
@@ -56,17 +62,48 @@ impl<C: FirebirdClient> SystemInfos for Connection<C> {
...
+        let version = conn.server_engine()?;
+
+        // Our current CI versions..
+        assert!([EngineVersion::V2, EngineVersion::V3, EngineVersion::V4].contains(&version));
+
+        Ok(())
+    }
+
+    #[test]
+    fn eng_version() {
```
dancingbug, line 104: if you use derive i think this is not necessary

src/utils.rs
```diff
@@ -0,0 +1,109 @@
...
+impl From<EngineVersion> for u8 {
+    fn from(eg: EngineVersion) -> Self {
+        match eg {
+            EngineVersion::V1 => 1,
+            EngineVersion::V2 => 2,
+            EngineVersion::V3 => 3,
+            EngineVersion::V4 => 4
+        }
+    }
+}
```
jairinhohw: You can define like this:

```suggestion
#[derive(Debug, Copy, Clone)]
#[repr(u8)]
pub enum EngineVersion {
    V1 = 1,
    V2 = 2,
    V3 = 3,
    V4 = 4,
}
```

And then convert to `u8` with `EngineVersion::V2 as u8`

src/utils.rs
jairinhohw: Just to avoid panics, it's better to use `&version.get(0..2)` in the match instead:

```rust
            return match &version.get(0..2) {
                Some("4.") => Ok(EngineVersion::V4),
                Some("3.") => Ok(EngineVersion::V3),
                Some("2.") => Ok(EngineVersion::V2),
                _ => Err(FbError::from(format!("Version not detected: {}", version)))
```

src/utils.rs
fernandobatels: I tried this before and it didn't work. But now, hilariously, it's working. I will fix

src/utils.rs
fernandobatels: Done!

src/utils.rs
fernandobatels: Nice. I will fix

src/utils.rs
fernandobatels: Done!

dancingbug (approved):

jairinhohw (approved):

PR #101, opened 2021-02-26
Diesel support base

fernandobatels:
Changes to support the diesel implementation.

From https://github.com/fernandobatels/rsfbclient/pull/86#issuecomment-786434843


rsfbclient-rust/src/client.rs
```diff
@@ -455,7 +459,18 @@ impl FirebirdWireConnection {
...
+        // Read lazy responses
+        for _ in 0..self.lazy_count {
+            if op_code != WireOp::Response as u32 {
+                return err_conn_rejected(op_code);
+            }
+            self.lazy_count -= 1;
+            parse_response(&mut resp)?;
+
+            op_code = resp.get_u32()?;
+        }
```
dancingbug, line 473: @jairinhohw I think you probably wrote this,

I think I understand the idea, but maybe you could elaborate a little bit the on 'lazy response' changes?

1. What's the intention? Is it to batch syscalls as much as possible and defer blocking?
2. Is wire protocol response order guaranteed to match TCP order of requests (if so- is that written somewhere or based just on observations?)
3. (followup to (2)) is there any other facility like a 'correlation id' to match req-resp pairs perfectly? it does not seem so.

any other details you think might be enlightening are also appreciated

jairinhohw: The link in the mailing list is indeed about the feature that makes it necessary to add those lazy responses verifications.

Some requests in the protocol version 11 and above do not cause the server to send the response immediately, the response is only sent after another request is made and needs a immediate response (does not have a deferred response), so all pending responses are sent from the server in the order of the requests, ending with the response to the last request the client made.

I've commented in the code the requests that causes deferred responses and incremented the `lazy_count` field to track how many there are, and in all parts that receive a server response and may have some pending deferred response to receive, I've added that check to receive any pending response.

About the order of the responses, at least up to protocol version 13 (i don't know the ones above this), there is no identifier to tie a request to a response, so all responses are sent in the same order of the requests.

Without this change, there was a chance that a response from a `free_statement` request was read in place of a response from another request, and the response of that request would be ignored.

This behavior is commented [here](https://firebirdsql.org/file/documentation/html/en/firebirddocs/wireprotocol/firebird-wire-protocol.html#wireprotocol-statements-free-v11).

src/connection/mod.rs
```diff
@@ -114,49 +126,129 @@ impl<C: FirebirdClient> Connection<C> {
...
+            tr.rollback(self).ok();
+        }
+
+        res
+    }
+
+    /// Run a closure with the default transaction, no rollback or commit will be automatically performed
+    /// after the closure returns. The next call to this function will use the same transaction
+    /// if it was not closed with `commit_retaining` or `rollback_retaining`
+    fn use_transaction<T, F>(&mut self, closure: F) -> Result<T, FbError>
```
dancingbug, line 180:
```suggestion
    pub fn use_transaction<T, F>(&mut self, closure: F) -> Result<T, FbError>
```

should it be `pub`?

fernandobatels: I think it shouldn't be. To me, this method is for internal utilization. 

Do you have any example where we need this method in the outside?

dancingbug: not really, just wondered why it had a doc comment in this case and wondered if pub was omitted by mistake

dancingbug (approved):
I  think the changes in the main crate involving `TransactionData` and the implementation of default transaction are not good, but the fundamental reason is because they are on top of a flawed design (for example duplication between user structures and `xyzData` structures)  so it also does not make much sense to me to block since it would be asking for radical changes making it difficult to move forward.

I also don't really feel so sure (as in i just don't know) about the changes in pure rust driver. I do think the `lazy_count` is probably sufficient for dealing with lazy/deferred messages (when message count = 1), but I'm not totally 100% sure. This is based on my trying to figure out FB source code over last few days. I haven't looked much at other driver implementations.

From looking at FB source I have a better understanding now of the architecture but many smaller implementation details are still not clear to me. For us the relevant parts are in `remote/` subdir ([link](https://github.com/FirebirdSQL/firebird/tree/master/src/remote)) particularly `remote/client/interface.cpp` (hint: means implementation of AP(Interface) classes in `include/firebird` headers)

In their implementation they use all kinds of queue/link structures to track individual queued/deferred messages instead of just keeping a count, (i found it most helpful to piece together understanding by looking at usage of main structs and their subparts: `rem_port` in `remote/remote.h` and `PACKET` in `remote/protocol.h`. `rmtque` is an important example of one of their queues).
Part of it seems to be for dealing with batched fetches properly (message count > 1) which won't be an issue in `pure_rust` for now, and maybe also dealing with multiple databases over a single connection? I still cannot really tell if that tracking of individual messages is necessary for any other reasons, for instance properly handling errors that occur in the middle of a stream.

For an example of these structures being used (which i still haven't totally understood): follow the flow of calls in `remote/client/interface.cpp` starting at `receive_packet` line inside `Statement::execute`.
hints: `rmtque_function` will have been set to be either `batch_dsql_fetch` or `batch_gds_receive` for each queue entry, and `dequeue_/enqueue_receive` mutate the `rmtque` structure to move forward or add new nodes. for TCP/IP the implementation will be using function pointers set in `inet.cpp` and via `remote.cpp:rem_port` method impls 

So although i am giving approval, my attitude is sort of "if it works for you it works for me". I see no _obvious_ implementation bugs at least.

jairinhohw (approved):

Issue #125, opened 2022-07-28
Fix r2d2 crate


fernandobatels:
Fix de r2d2 build

_Originally posted by @nt202 in https://github.com/fernandobatels/rsfbclient/issues/123#issuecomment-1198181600_

fernandobatels:
@nt202.

The [r2d2](https://crates.io/crates/r2d2_firebird) with `v0.19.0` is available now.

nt202:
@fernandobatels, Thank you very much! :)

PR #126, opened 2022-07-29
Crate r2d2 build fixed

fernandobatels:
Fixed the release build of r2d2_firebird crate. Now we have a phantom option on simple connection/transaction enum's.

Without this fix, we cannot build the r2d2 on release mode.

Issue #125

### DIFF
--- a/src/connection/simple.rs
+++ b/src/connection/simple.rs
@@ -27,6 +27,12 @@ pub(crate) enum TypeConnectionContainer {
     NativeDynLoad(Connection<NativeFbClient<DynLoad>>),
     #[cfg(feature = "pure_rust")]
     PureRust(Connection<RustFbClient>),
+    #[cfg(all(
+        not(feature = "pure_rust"),
+        not(feature = "linking"),
+        not(feature = "dynamic_loading")
+    ))]
+    Phantom,
 }
 
 #[cfg(feature = "linking")]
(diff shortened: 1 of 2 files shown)


fernandobatels:
> I don't understand the problem, the crate can be built as long as one of the 3 main rsfbclient features are enables (`linking`, `dynamic_loading`, or `pure_rust`). If we merge this, it will allways pull the `pure_rust` implementation, even if not necessary (using `linking` for example).

On `master` branch, running `cargo build --release` inside `rd2d_firebird`:
```
error[E0392]: parameter `'c` is never used
  --> src/transaction/simple.rs:19:30
   |
19 | pub struct SimpleTransaction<'c> {
   |                              ^^ unused parameter
```

The `cargo publish` build the crate on release mode. Without my fix, I cannot launch new versions of `rd2d`.

I agree with your commentary about always pull the `pure_rust`. I will find an alternative.

jairinhohw:
Does it work if you do `cargo publish --features rsfbclient/pure_rust` ?

fernandobatels:
> Does it work if you do `cargo publish --features rsfbclient/pure_rust` ?

No. Same error.

fernandobatels:
> cargo publish --features rsfbclient/pure_rust

It worked. Thanks. This PR is no longer needed.

Issue #135, opened 2022-09-27
Missing important transaction parameters


datiscum:
When testing various "select" statements, the test programme suddenly stopped and did not return. The problem was an open transaction in another database connection.
Immediately after committing the open transaction, the test programme returned the corresponding values.
When I looked for the problem in rsfbclient, I noticed that it only knows one parameter for the configuration of the transactions.

pub enum TrIsolationLevel {
    /// Transactions can't see alterations committed after they started
    Concurrency = ibase::isc_tpb_concurrency as u8,
    /// Table locking
    Concistency = ibase::isc_tpb_consistency as u8,
    /// Transactions can see alterations committed after they started
    ReadCommitted = ibase::isc_tpb_read_committed as u8,}

Here all other important transaction parameters are missing to solve exactly this problem.  

https://firebirdsql.org/file/documentation/html/en/refdocs/fblangref30/firebird-30-language-reference.html#fblangref30-transacs-settransac

Transaction Parameters

The main parameters of a transaction are:
    data access mode (READ WRITE, READ ONLY)
    lock resolution mode (WAIT, NO WAIT) with an optional LOCK TIMEOUT specification
    isolation level (READ COMMITTED, SNAPSHOT, SNAPSHOT TABLE STABILITY).

My standard in GUI applications for Firebird is:
nowait
read_committed
rec_version

fernandobatels:
I will check this tonight. Anyway, do you can provide a code example of your program?

> The problem was an open transaction in another database connection.

Was this second connection made by your process? Or by another program like flamerobin?

datiscum:
> I will check this tonight. Anyway, do you can provide a code example of your program?
I use the supplied "select.rs" example.
Only read a few values from the DB with select.

> 
> > The problem was an open transaction in another database connection.
> 
> Was this second connection made by your process? Or by another program like flamerobin?
flamerobin yes! 
I forgot to confirm the change in the table used by "select" and then the "select.rs" got stuck, which then pointed me to the problem.

I would like to be able to specify the transaction parameters that are to be used by default when connecting to the database.
And in self-started transactions, the parameters can then also be passed again.

fernandobatels:
> I will check this tonight. 

The problem is real. 

> I would like to be able to specify the transaction parameters that are to be used by default when connecting to the database.
And in self-started transactions, the parameters can then also be passed again when needed.

I will start the study to implement this feature. Maybe in a faw days it will be ready for testing.

[...]

PR #136, opened 2022-10-09
Transaction configuration

fernandobatels:
Now we can customize the transaction behavior of our crate. 

Like others builder options, we just need to call the `with_transaction`. With this method the user can access the `TransactionConfigurationBuilder` and setup what they need.

This method setup the default configuration, but the user also can choose a new configuration for a specific transaction.

### DIFF
--- a/src/connection/mod.rs
+++ b/src/connection/mod.rs
@@ -3,7 +3,7 @@
 //!
 //! Connection functions
 //!
-use rsfbclient_core::{Dialect, FbError, FirebirdClient, FirebirdClientDbOps, FromRow, IntoParams};
+use rsfbclient_core::{Dialect, FbError, FirebirdClient, FirebirdClientDbOps, FromRow, IntoParams, TransactionConfiguration};
 use std::{marker, mem};
 
 use crate::{
@@ -170,7 +170,7 @@ impl<C: FirebirdClient> Connection<C> {
         let mut tr = if let Some(tr) = self.def_tr.take() {
             tr.into_transaction(self)
         } else {
-            Transaction::new(self)?
+            Transaction::new(self, TransactionConfiguration::default())?
         };
 
         let res = closure(&mut tr);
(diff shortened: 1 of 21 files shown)


src/connection/mod.rs
```diff
@@ -161,7 +171,11 @@ impl<C: FirebirdClient> Connection<C> {
 
     /// Run a closure with a transaction, if the closure returns an error
     /// and the default transaction is not active, the transaction will rollback, else it will be committed
-    pub fn with_transaction<T, F>(&mut self, closure: F) -> Result<T, FbError>
+    pub fn with_transaction<T, F>(
+        &mut self,
+        custom_confs: Option<TransactionConfiguration>,
+        closure: F,
+    ) -> Result<T, FbError>
```
jairinhohw, line 178: I would prefer the `with_transaction` to use the default transaction configuration and add a `with_transaction_config` or similar that you can pass the `TransactionConfiguration` option,
so the user does not need to pas a `None`or `Some`.

fernandobatels: Done. Looks better.

jairinhohw (approved):

fernandobatels:
@datiscum can you check this PR? Does the implementation solve your problem in #135 ?

datiscum:
> 

Yes, thank you!
If I now set the transaction parameters as they are set in the GUI app, the query no longer hangs.
My connect:

    let mut conn = rsfbclient::builder_native()
        .with_dyn_link()
        .with_remote()
        .host("192.168.11.22")
        .db_name("/DB/EventTest.fdb")
        .user("SYSDBA")
        .pass("!master?")
        .transaction(TransactionConfiguration {
          lock_resolution: TrLockResolution::NoWait,
          isolation: TrIsolationLevel::ReadCommited(TrRecordVersion::RecordVersion),
          data_access: TrDataAccessMode::ReadWrite
      })
        .connect()?;

PR #142, opened 2023-01-25
Firebird events support

fernandobatels:
Firebird events support added. Now the users can listen for posted events by triggers and others.

Only for native client. The wire protocol is too complex to implement now.

Basically now we have `wait_for_event` method directly on the connection struct and a more complex solution with `listen_event` method from `RemoteEventManager`.

### DIFF
--- a/rsfbclient-native/src/connection.rs
+++ b/rsfbclient-native/src/connection.rs
@@ -586,6 +586,50 @@ impl<T: LinkageMarker> FirebirdClientSqlOps for NativeFbClient<T> {
 
         Ok(rcol)
     }
+
+    /// Register a listner for some events
+    fn que_events(
+        &mut self,
+        db_handle: &mut Self::DbHandle,
+        names: Vec<String>
+    ) -> Result<(), FbError> {
+
+        let mut event_buffer: Vec<u8> = Vec::with_capacity(256);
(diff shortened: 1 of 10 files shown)


rsfbclient-native/src/connection.rs
```diff
@@ -588,6 +590,71 @@ impl<T: LinkageMarker> FirebirdClientSqlOps for NativeFbClient<T> {
...
 }
 
+impl<T: LinkageMarker> FirebirdClientDbEvents for NativeFbClient<T> {
+    fn wait_for_event(
+        &mut self,
+        db_handle: &mut Self::DbHandle,
+        name: String,
+    ) -> Result<(), FbError> {
+        let mut event_buffer = ptr::null_mut();
+        let mut result_buffer = ptr::null_mut();
```
jairinhohw, line 600: It seems these buffers need to be freed with `isc_free` before the end of the function, as they are allocated by the `isc_event_block` function.
https://github.com/FirebirdSQL/php-firebird/blob/b6c288326678f7b8613f5a7d0648ad010c67674c/ibase_events.c#L45

jairinhohw (changes requested):

rsfbclient-native/src/connection.rs
fernandobatels: Solved

jairinhohw (approved):

PR #152, opened 2024-01-05
Dialect on create_database and no_db_trigger support

fernandobatels:
The `dialect` connection builder options now are also applied on database creation, provided by `create_database` method. 

`no_db_trigger` option added on connection builder, applied on database connection, disabling the database triggers. Table triggers are no affected.

### DIFF
--- a/rsfbclient-core/src/connection.rs
+++ b/rsfbclient-core/src/connection.rs
@@ -37,6 +37,7 @@ pub trait FirebirdClientDbOps: Send {
         &mut self,
         config: &Self::AttachmentConfig,
         dialect: Dialect,
+        no_db_triggers: bool,
     ) -> Result<Self::DbHandle, FbError>;
 
     /// Disconnect from the database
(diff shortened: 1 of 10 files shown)


rsfbclient-core/src/connection.rs
```diff
@@ -145,10 +148,11 @@ pub trait FirebirdClientDbEvents: FirebirdClientDbOps {
...
 
-#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Default)]
 #[repr(u8)]
 /// Firebird sql dialect
 pub enum Dialect {
+    #[default]
     D1 = 1,
     D2 = 2,
     D3 = 3,
```
jairinhohw, line 158: I think dialect 3 should be the default one, as it is the default for the connection

```suggestion
#[derive(Debug, Eq, PartialEq, Copy, Clone, Default)]
#[repr(u8)]
/// Firebird sql dialect
pub enum Dialect {
    D1 = 1,
    D2 = 2,
    #[default]
    D3 = 3,
```

PR #158, opened 2024-08-22
fix compatibility issues 

Noriebalbinot:
using slice::from_raw_parts to genereta a new slice with a mutable size. to avoid the new bound checking on get_unchecked and get_unchecked_mut


fernandobatels (approved):

fernandobatels:
@Noriebalbinot new 0.25.0 released with this changes

PR #171, opened 2026-07-01
pure_rust: batch op_fetch (fetch N rows per round-trip)

arthurbolsoni:
## Problem

The pure-Rust backend fetches one row per `op_fetch`, i.e. one TCP round-trip
per row (`wire.rs`: `put_u32(1) // TODO: increase to return more rows in one
fetch request`). Over a WAN link this caps throughput at the round-trip rate
(~170 rows/s at ~5 ms RTT) regardless of CPU — the native (C client) backend,
which batches fetches, is ~30x faster on the same link.

## Change

- `op_fetch` now requests `FB_FETCH_BATCH` rows per round-trip (env var, default
  200). Rows are served from a per-statement buffer, so `fetch()` still returns
  one row at a time — streaming semantics and bounded memory are unchanged.
- Robust reading: `op_fetch_response` packets can be split across TCP segments.
  The batch reader parses greedily and, on a genuine byte-underflow, reads more
  and resumes. A real data error (e.g. invalid UTF-8 when decoding a column)
  still propagates instead of looping forever.
- The batch state (`prefetched`, `cursor_eof`) is reset on (re)execute, so
  re-running a prepared statement correctly reopens its cursor.

No public API change; default behaviour is just faster. `FB_FETCH_BATCH=1`
reproduces the old one-row-per-round-trip behaviour.

## Performance

187k-row query, same link:

| backend | throughput | CPU |
|---|---|---|
| pure_rust (before) | ~170 rows/s | round-trip bound |
| pure_rust (after, batch 200) | ~5000 rows/s | 5.1 s |
| native (C client) | ~5350 rows/s | 8.1 s |

After batching the pure-Rust backend matches the native client's throughput and
uses less CPU (no FFI/copy overhead).

## Tests

`cargo test --no-default-features --features pure_rust -- --test-threads 1`: all
green, including a new regression test `row::fetch_batches_crossing_boundary`
that fetches 500 rows to cross batch boundaries and asserts count and order.
Also verified with `FB_FETCH_BATCH` = 1, 7, 10, 200.

### DIFF
--- a/rsfbclient-rust/src/client.rs
+++ b/rsfbclient-rust/src/client.rs
@@ -2,6 +2,7 @@
 
 use bytes::{BufMut, Bytes, BytesMut};
 use std::{
+    collections::VecDeque,
     env,
     io::{Read, Write},
     net::TcpStream,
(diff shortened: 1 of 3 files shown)


fernandobatels (approved):

fernandobatels:
Thanks! @arthurbolsoni 

Here’s the new version: https://github.com/fernandobatels/rsfbclient/releases/tag/v0.27.0

PR #173, opened 2026-07-21
Add isolation/type-mapping examples and document transaction and type semantics

mariuz:
Hi! We recently ported the 43 hands-on companion samples of a Firebird architecture paper to rsfbclient ([samples/rust here](https://github.com/mariuz/conceptual-architecture-for-firebird-paper/tree/master/samples/rust)) — every subsystem document now has a verified rsfbclient twin next to its C++/fb-cpp/node-firebird siblings. Along the way we collected the behaviors that surprised us, and this PR feeds the two most useful ones back as examples + documentation, per CONTRIBUTING's invitation for docs carrying Firebird knowledge.

## New examples

**`examples/isolation_levels.rs`** — explicit `SimpleTransaction` objects with `TransactionConfiguration`: SNAPSHOT vs READ COMMITTED visibility across two connections, and a `NoWait` update conflict surfacing the `-913` chain with the concurrent transaction number. The header explains *why* explicit transactions are the right tool here: `Connection::commit`/`rollback` are retaining, so the default transaction keeps its first configuration — and, under SNAPSHOT, its first snapshot — for the life of the connection. (Our first port used the default transaction with a "NoWait" config passed to `begin_transaction_config` after the first statement, and watched the update block forever under the connect-time WAIT default. Documented, in hindsight — but easy to fall into, so now it's an example.)

**`examples/type_mapping.rs`** — every supported `SqlType` mapping fetched typed, plus the two sharp edges shown live: `NUMERIC(18,2)` storing `90071992547409.93` reads back `.92` through `f64` (scaled integer past 2^53), exact via `CAST`; and INT128/DECFLOAT failing at describe time with `Unsupported column type (32752/32762)`, with the CAST-to-VARCHAR route delivering all digits. The FB4 columns are added in a separate allowed-to-fail DDL step so the example still runs against Firebird 3.

## Extended documentation

Two new chapters in the `src/lib.rs` front-page doc, between "Simple Connection/Transaction" and "Cargo features":

- **Transactions** — the default transaction's lifecycle (lazy start with the connect-time configuration, retaining commit/rollback, configuration fixed at first use, `begin_transaction_config` reusing the existing physical transaction), and when to reach for explicit `Transaction`/`SimpleTransaction` objects.
- **Data type mappings** — the supported conversions as a table, plus the f64-precision and FB4-types caveats with their workarounds.

## Verification

- Both examples run verified against Firebird 6.0 (`LI-T6.0.0.2076`, Linux aarch64, `linking` feature): snapshot holds at 100 while the other connection commits 150; the NoWait conflict raises `-913 update conflicts with concurrent update`; the INT128/DECFLOAT paths produce exactly the documented errors and CAST results.
- `cargo build --examples` clean (no new warnings), `cargo fmt --all -- --check` passes, `cargo doc` adds no new warnings.

Happy to adjust scope or wording — and if you'd like, we can follow up with issues for the sharp edges we hit that this PR only documents (e.g. `EngineVersion` stopping at V5 makes `SystemInfos::server_engine()` fail on Firebird 6).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/isolation_levels.rs
+++ b/examples/isolation_levels.rs
@@ -1,0 +1,137 @@
+//!
+//! Rust Firebird Client
+//!
+//! Example of explicit transactions and isolation levels
+//!
+//! Two connections to the same database demonstrate what the
+//! TransactionConfiguration options actually change inside Firebird:
+//!
+//!  1. Visibility: a Concurrency (SNAPSHOT) transaction does not see a
+//!     commit that happens after it starts; a ReadCommited one does.
+//!  2. Conflict policy: with TrLockResolution::NoWait, updating a row
+//!     that another transaction updated first fails immediately with
+//!     "update conflicts with concurrent update" instead of blocking.
+//!
+//! The example uses explicit `SimpleTransaction` objects instead of the
+//! connection's default transaction on purpose: the default transaction
+//! is started once with the configuration given at connect time, and
+//! `Connection::commit`/`rollback` are RETAINING operations — the same
+//! physical transaction (with its configuration and, under SNAPSHOT,
+//! its snapshot) lives for the whole life of the connection. Explicit
+//! transaction objects give each demonstration its own configuration
+//! and a real, consuming commit.
+//!
+//! The table is created by the example itself (idempotent); you only
+//! need an `examples.fdb` database.
+//!
+
+#![allow(unused_variables, unused_mut)]
+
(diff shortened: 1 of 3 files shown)


fernandobatels (approved):

fernandobatels:

PR #176, opened 2026-07-29
enable TCP_NODELAY on socket to speed up pure rust client

MnlPhlp:
This is an optimization for the pure rust client i found with some help of claude.

In my use case it shortened many requests by an oder of magnitude or more just because the socket does not wait for a delayed ACK.

## Measurements
| statement | before | after |
|---|---|---|
| `SELECT 1 FROM RDB$DATABASE` | 44 ms | 0.17 ms |
| column metadata query (`RDB$RELATION_FIELDS` join) | 86 ms | 1.8 ms |
| indexed select returning no rows | 84 ms | 2.1 ms |
| 1000-row indexed select | 107 ms | 5.7 ms |

### DIFF
--- a/rsfbclient-rust/src/client.rs
+++ b/rsfbclient-rust/src/client.rs
@@ -323,6 +323,11 @@ impl FirebirdWireConnection {
         charset: Charset,
     ) -> Result<Self, FbError> {
         let socket = TcpStream::connect((host, port))?;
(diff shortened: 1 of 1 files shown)


fernandobatels (approved):

fernandobatels:

PR #177, opened 2026-08-01
Modified `pure_rust` to work with WireCrypt=Disabled

### LINKED ISSUE #172: Pure Rust client: connection fails with “unavailable database” when the server has WireCrypt = Disabled (op_accept_data handshake path not implemented)
## Summary

With the `pure_rust` feature, `connect()` fails with `Sql { msg: "unavailable database", code: -1 }` against any Firebird 3/4 server configured with `WireCrypt = Disabled`. The same build connects fine when the server has `WireCrypt = Enabled` or `Required`. `isql` connects fine in all cases, so it is client-side.

`WireCrypt = Disabled` is a common setting in shops that still run old fbclient-based apps alongside newer tooling, so this makes the pure Rust client unusable against a lot of real-world servers.

## Environment

- rsfbclient 0.27.0 (current release), `default-features = false, features = ["pure_rust"]`
- Firebird 3.0.14 x64 on Linux (also reproduced earlier against Firebird 4.0.7)
- Rust 1.96

## Minimal repro

```toml
[dependencies]
rsfbclient = { version = "0.27.0", default-features = false, features = ["pure_rust"] }
```

```rust
fn main() {
    let res = rsfbclient::builder_pure_rust()
        .host("127.0.0.1")
        .port(3050)
        .db_name("examples") // any alias/path
        .user("SYSDBA")
        .pass("masterkey")
        .connect();
    match res {
        Ok(_) => println!("CONNECTED OK"),
        Err(e) => println!("CONNECT ERROR: {:?}", e),
    }
}
```

Server side, in `firebird.conf`:

```
AuthServer = Srp
WireCrypt = Disabled
```

## Results across server configs (same client build, same credentials)

| AuthServer | WireCrypt | Result |
|---|---|---|
| Srp | Disabled | `CONNECT ERROR: Sql { msg: "unavailable database", code: -1 }` |
| Srp | Enabled | CONNECTED OK |
| Srp | Required | CONNECTED OK |
| Srp, Legacy_Auth | Disabled | `CONNECT ERROR: Sql { msg: "unavailable database", code: -1 }` |
| Srp, Legacy_Auth | Enabled | CONNECTED OK |
| Srp, Legacy_Auth | Required | CONNECTED OK |

## Where it goes wrong

I instrumented the handshake. The difference is the accept opcode the server chooses:

- `WireCrypt = Enabled/Required`: server replies **op_cond_accept (98)**. The client continues auth with `op_cont_auth`, the SRP proof is accepted, `op_crypt` follows, connection succeeds.
- `WireCrypt = Disabled`: server replies **op_accept_data (94)** with the same Srp salt/public key and `authenticated = false`. The client sends the proof via `op_cont_auth` exactly as in the other case, and the server answers with `isc_unavailable` ("unavailable database").

`parse_accept` treats `op_accept_data` and `op_cond_accept` identically, and `Connection::connect` always continues via `op_cont_auth` (`srp_auth` in `rsfbclient-rust/src/client.rs`). But per the v13 wire protocol, `op_cont_auth` is only valid after `op_cond_accept`. After `op_accept_data` the client is expected to carry the remaining auth data inside the following request, i.e. put the SRP proof in the attach DPB as `isc_dpb_specific_auth_data` (this is what Jaybird and the .NET provider do). When wire encryption is possible the server has to finish auth up front to establish the session key, hence `op_cond_accept`; with crypt disabled it defers auth into the attach, hence `op_accept_data`. The pure Rust client only implements the first path.

A secondary issue in the same function: `srp_auth` sends `op_crypt("Arc4", "Symmetric")` unconditionally after auth. That happens to work today because the broken `WireCrypt = Disabled` path never reaches it, but once `op_accept_data` is handled, starting Arc4 against a server that declined wire encryption will desync the stream. The crypt step should be conditional on the negotiated result.

## Suggested fix

Handle `op_accept_data` by completing SRP as part of the attach request (`isc_dpb_specific_auth_data`, plugin name and list in the DPB), and only send `op_crypt` when encryption was actually negotiated.

## Workaround

Removing the V13 entry from the proposed protocol list in `wire.rs` makes the client fall back to protocol 12 with legacy password auth, which connects fine against servers that also offer `Legacy_Auth`. That is what we shipped internally.

## Offer: two small hardening patches

We vendored the crate for a while and carried two more small changes. Happy to open separate issues/PRs for either if there is interest:

1. **Handshake read timeout** (`client.rs`): set a read timeout on the socket for the connect/handshake phase and clear it once connected, so a stalled negotiation surfaces as an error instead of blocking forever. We observed one indefinite hang during early testing (protocol 12 client against an SRP-only server) that we can no longer reproduce against 3.0.14, so consider this cheap insurance rather than a verified bug fix.
2. **Opt-in lenient charset decode** (`rsfbclient-core/src/charset.rs`): legacy databases sometimes contain bytes that violate their declared charset, and today a single bad byte makes the whole row unreadable. We swapped `DecoderTrap::Strict` for `Replace` (and `from_utf8_lossy` for the no-charset case) in our fork. As a default that silently changes behaviour, so upstream it should probably be a builder option (strict by default). Worth it for tools that browse old data.

Full disclosure: we have since migrated the project that hit this to the official .NET provider, so we are offering the diagnosis and patches as-is rather than as something we depend on. The repro above is self-contained and I am happy to answer questions about it.

jairinhohw:
Fixes #172

### DIFF
--- a/rsfbclient-rust/src/client.rs
+++ b/rsfbclient-rust/src/client.rs
@@ -83,6 +83,11 @@ pub struct FirebirdWireConnection {
     lazy_count: u32,
 
     pub(crate) charset: Charset,
+
+    /// AuthPlugin data for use on attach when WireCrypt = Disabled
+    pub(crate) auth_plugin: Option<AuthPlugin>,
+    /// Key for the srp auth
+    pub(crate) srp_key: [u8; 32],
 }
 
 /// Data to keep track about a prepared statement
@@ -349,17 +354,22 @@ impl FirebirdWireConnection {
 
         let ConnectionResponse {
             version,
-            auth_plugin,
+            mut auth_plugin,
+            continue_auth,
         } = parse_accept(&mut resp)?;
 
-        if let Some(mut auth_plugin) = auth_plugin {
+        if let Some(auth_plugin) = &mut auth_plugin {
             loop {
                 match auth_plugin.kind {
                     plugin @ AuthPluginType::Srp => {
                         let srp = SrpClient::<sha1::Sha1>::new(&srp_key, &SRP_GROUP);
 
-                        if let Some(data) = auth_plugin.data {
-                            socket = srp_auth(socket, &mut buff, srp, plugin, user, pass, data)?;
+                        if let Some(data) = auth_plugin.data.clone() {
+                            if continue_auth {
+                                // Continue autentication if needed
+                                socket =
+                                    srp_auth(socket, &mut buff, srp, plugin, user, pass, &data)?;
+                            }
 
                             // Authentication Ok
                             break;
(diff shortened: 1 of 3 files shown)


fernandobatels (approved):

fernandobatels:

PR #179, opened 2026-08-06
fix(pure_rust): read responses until complete, keep the surplus

d0m1n1kr:
## Problem

The Firebird wire protocol is a byte stream without packet framing, but the
pure-Rust client's `read_packet` treats one `socket.read()` as one response: it
only reads again when the read has exactly filled its 2 KB buffer. Two failure
modes follow:

1. **A response split across TCP segments** reaches the parser truncated —
   `"Invalid server response, missing bytes"` / `"Invalid Xsqlda received from
   server"` — and its remainder stays in the socket, so every later operation
   reads leftover bytes as its op code: `"Connection rejected with code 0"`
   (or `4294967295`, or any other garbage value). Wide tables hit this
   routinely — their prepare/XSQLDA response alone is easily over 2 KB.
2. **Two responses coalescing into one read** lose the second one, with the
   same result.

Once misaligned the connection never recovers, and the failure is worse than
the error messages suggest: a misaligned stream can answer the *current*
statement with the metadata or rows of the *previous* one — plausible-looking
wrong data instead of an error.

On current `master` this is easy to hit: against a stock
`firebirdsql/firebird:3.0.11` container (SRP + WireCrypt), `pure_rust` fails
nondeterministically during the handshake or on the first multi-column query.
`v0.27.0` from crates.io shows the same failures.

## Fix

Responses are now read through `read_with()`, which parses the accumulated
bytes and reads more from the socket only when the parse underflows, then keeps
whatever the parse did not consume in a per-connection `pending` buffer for the
next call. That is the framing rule the protocol actually requires: a response
is complete exactly when it parses.

Two consequences of the same root cause are fixed along with it:

- The end-of-cursor `op_fetch_response` is `[status=100][count]`, but
  `parse_fetch_response` stops after the status. With the surplus no longer
  discarded, those four bytes would be read as the next op code, so
  `parse_one_fetch_response` now consumes both.
- `fetch_batch` resolved blob columns while later rows of the same `op_fetch`
  batch were still queued on the socket, so the blob round-trips read those
  queued rows instead of their own replies. Blobs are now resolved after the
  batch has been fully consumed; `execute2` does the same.

`lazy_count` consumption is rolled back when a parse attempt underflows, so
retries are idempotent.

## Evidence

Same workload (wide table, all text columns read via
`CAST(... AS BLOB SUB_TYPE 0)`, batched fetch, every value verified), same
stock Firebird 3.0.11 server:

| revision | result |
|---|---|
| `master` (c5ab7c3) | fails: `Connection rejected with code 1 (Connect)` |
| `master` + this PR | 2000 rows verified over repeated rounds, every run |

A self-contained reproduction lives in a companion repository —
**https://github.com/d0m1n1kr/rsfbclient-desync-repro** — with two pinned
builds of the identical workload (`broken/` = upstream master,
`fixed/` = this branch); `./run.sh` starts Firebird 3.0.11 in Docker and runs
both, expecting broken to fail and fixed to pass.

Also verified against Firebird 2.5 (plaintext handshake, no WireCrypt) on the
v0.27.0 base: unfixed fails on the first batched query, fixed verifies 5000
rows at both the default `FB_FETCH_BATCH` and `=1`.

Three unit regression tests cover the failure modes directly: a response
delivered one byte at a time, two responses arriving in a single read, and the
end-of-cursor trailer.

## Disclosure

This patch was developed with the assistance of Claude (Anthropic's Fable
model) — analysis, implementation and the reproduction harness. The commit
carries a `Co-Authored-By` trailer accordingly. It was reviewed by a human and
all results above come from runs against real Firebird servers (2.5 and
3.0.11), not from the model.

## Unrelated observation

While testing baselines: current `master` fails to attach to Firebird 2.5 with
`"Invalid clumplet buffer structure: buffer end before end of clumplet"` —
this predates this PR (reproducible on c5ab7c3 without it) and looks like a
DPB-encoding regression from the WireCrypt work. Happy to file it as a
separate issue.

### DIFF
--- a/rsfbclient-rust/src/client.rs
+++ b/rsfbclient-rust/src/client.rs
@@ -33,22 +33,17 @@ fn fetch_batch_size() -> u32 {
         .unwrap_or(200)
 }
 
-/// Result of parsing ONE op_fetch_response.
+/// Result of parsing ONE op_fetch_response. Blob columns are still unresolved:
+/// fetching them costs extra round-trips that must not be interleaved with the
+/// responses of the running batch (see `fetch_batch`).
 enum FetchOne {
     /// A row (status=0, messages=1).
-    Row(Vec<Column>),
+    Row(Vec<ParsedColumn>),
     /// End of THIS batch (status=0, messages=0): the server ended the op_fetch
     /// without exhausting the cursor. Re-issuing op_fetch fetches the rest.
     BatchEnd,
     /// End of cursor (status=100). Nothing more to read.
     End,
-}
-
-enum FetchErr {
-    /// Not enough bytes in the buffer — read more from the socket and retry.
-    NeedMore,
-    /// A real protocol/server error — propagate it.
-    Fatal(FbError),
 }
 
 /// Firebird client implemented in pure rust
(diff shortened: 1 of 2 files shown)


fernandobatels (approved):

fernandobatels:

PR #180, opened 2026-08-11
Implement firebird events in the pure rust client

maratth:
Firebird events in the pure rust client
wait_for_event and listen_event only worked with rsfbclient-native. With pure_rust they returned "Events only works with the native client".

This implements FirebirdClientDbEvents for RustFbClient: op_connect_request to open the auxiliary connection firebird delivers notifications on, then op_que_events / op_event / op_cancel_events. Event names are encoded with the connection charset, since the server matches them byte for byte against POST_EVENT.

No API change — the existing events API just starts working. rsfbclient-core and rsfbclient-native are untouched, and there is no new dependency.

Tests
29 unit tests covering the packet layouts, the event parameter block and the charset handling, none of which need a database. The existing integration tests now also run for pure_rust, plus two new ones.

Known limits
One event name per wait, and no occurrence count: the trait carries neither. 
The wait blocks the connection and can't be interrupted, same as the native client.

### DIFF
--- a/examples/events.rs
+++ b/examples/events.rs
@@ -9,47 +9,52 @@
 use rsfbclient::{FbError, Queryable, RemoteEventsManager};
 
 fn main() -> Result<(), FbError> {
-    #[cfg(not(feature = "pure_rust"))] // No support for events with pure rust driver
-    {
-        #[cfg(feature = "linking")]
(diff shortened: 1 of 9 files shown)


fernandobatels (approved):

fernandobatels:

